PR #23, opened 2015-12-21
Check that attr exists before access

amlweems:
Fixes the `AttributeError: LibVEX_ShowStats` error when importing pyvex with the latest version of cffi.


zardus:
Thanks, man!

Issue #47, opened 2016-10-31
Using pyvex for transformations/instrumentation


mattrepl:
Hello, currently pyvex makes a call into C code to create an IRSB. That IRSB and its components (IRStmt, etc) are exposed as Python objects, but there doesn't seem to be a way to create an empty IRSB or construct IRStmt instances not backed by guest machine instructions.

I'd like to modify IRSBs, IRStmts, and IRExprs generated from code as well as programmatically generating instances without backing code. Is this functionality within the scope of pyvex? I'm happy to work on it and submit a patch, if so.

rhelmot:
This is a longstanding TODO for pyvex, unfortunately... it shouldn't be any
harder than just shuffling around the constructors for all the IR objects,
but it'll be some work...

mattrepl:
Glad to hear it's a TODO.

There are a bunch of ways we can support creating IR component instances from the corresponding existing C structs (the current implementation) and from sub components (this is what the constructors in libvex look like).

We probably don't want to break the pyvex API at this point (maybe in major version bump?), so the current implementation of __init__ would remain. Each IR component class can also have a static method which mirrors the underlying libvex constructor. Eventually we could move the current __init__ to a static method and then move the libvex-style constructor in __init__.

Does adding support for libvex-style constructors via static methods attached to respective IR component classes sound reasonable to you all?

rhelmot:
Yes, that was exactly what I was thinking of! If you want to just do this,
a PR would be totally welcome :)

Unfortunately, we have to keep all the angr components version numbers in
lockstep otherwise dependency hell becomes a serious problem... One way to
keep everything totally compatible would be to have there only be one
constructor, but you can pass in either a raw struct pointer as a keyword
argument or the normal construction parameters. Might be a little harder to
set up, though.

mattrepl:
Ok, I'm going forward with a `create` staticmethod attached to all the IR component classes (we can change the name to whatever) that will mirror the libvex constructors.

Is it OK for us to use "ABI mode" with cffi for calling the libvex constructors? It looks like you all wrote your own wrapper (pyvex.c) which gets statically linked with libvex. Since all the libvex functions, structs, etc are included in the generated vex_ffi.py module we can reference them without making any changes.

It sounds like the preferred approach is to include the current contents of pyvex.c with a C extension module generated by cffi (see "API mode"). I'm leaning towards going ahead and leaving pyvex.c alone and referencing the constructors via ABI mode, at least for now.

[...]

PR #51, opened 2016-12-09
IR component creation in Python

mattrepl:
Here's the final commit of changes to allow Python creation of IR components without converting from the corresponding C structs. The initial discussion is in issue #47.

I kept the correspondence of fields between the C structs and the Python classes 1:1. I.e., you can roundtrip a Python IR instance to C and back to Python and no information is lost. Previously, Python IR instances had `result_size` and `result_type` fields, those have been converted to methods that take a type environment parameter. I searched around and the only external code that refers to those fields was in simuvex. I made the changes in simuvex to support this here: https://github.com/mattrepl/simuvex/tree/use_irsb_tyenv

I also changed how printing of Put statements and Get expressions with architecture register names is done. It's a little ugly, but it prevents needing the Python IR instances needing a type environment to be instantiated. Note that the IR C structs do not get coupled with a type environment; so pyvex now more closely follows libvex. I don't see a problem with this, other than the hack to get nice register names when printing statements and expressions.

Have started updating the unit tests as well, to make sure nothing is broken from these changes. That'll come in a later commit.

### DIFF
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -81,9 +81,9 @@ class IRSB(VEXObject):
 
             self.c_irsb = c_irsb
             self.arch = arch
-            self.tyenv = IRTypeEnv(c_irsb.tyenv)
-            self.statements = [stmt.IRStmt._translate(c_irsb.stmts[i]) for i in xrange(c_irsb.stmts_used)]
-            self.next = expr.IRExpr._translate(c_irsb.next)
+            self.tyenv = IRTypeEnv._from_c(c_irsb.tyenv)
+            self.statements = [stmt.IRStmt._from_c(c_irsb.stmts[i]) for i in xrange(c_irsb.stmts_used)]
+            self.next = expr.IRExpr._from_c(c_irsb.next)
             self.offsIP = c_irsb.offsIP
             self.stmts_used = c_irsb.stmts_used
             self.jumpkind = ints_to_enums[c_irsb.jumpkind]
@@ -265,14 +265,26 @@ class IRTypeEnv(VEXObject):
 
     __slots__ = [ 'types', 'types_used' ]
 
-    def __init__(self, tyenv):
+    def __init__(self, types, types_used):
         VEXObject.__init__(self)
-        self.types = [ints_to_enums[tyenv.types[t]] for t in xrange(tyenv.types_used)]
-        self.types_used = tyenv.types_used
+        self.types = types
+        self.types_used = types_used
 
     def __str__(self):
         return ' '.join(("t%d:%s" % (i, t)) for i, t in enumerate(self.types))
 
+    @staticmethod
+    def _from_c(c_tyenv):
+        return IRTypeEnv([ints_to_enums[c_tyenv.types[t]] for t in xrange(c_tyenv.types_used)],
+                         c_tyenv.types_used)
+
+    @staticmethod
+    def _to_c(tyenv):
+        c_tyenv = pvc.emptyIRTypeEnv()
+        for ty in tyenv.types:
+            pvc.newIRTemp(c_tyenv, enums_to_ints[ty])
+        return c_tyenv
+
 from . import expr, stmt, ffi, pvc
-from .enums import ints_to_enums
+from .enums import ints_to_enums, enums_to_ints
 from .errors import PyVEXError
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -82,8 +82,8 @@ class IRSB(VEXObject):
             self.c_irsb = c_irsb
             self.arch = arch
             self.tyenv = IRTypeEnv(c_irsb.tyenv)
-            self.statements = [stmt.IRStmt._translate(c_irsb.stmts[i], self) for i in xrange(c_irsb.stmts_used)]
-            self.next = expr.IRExpr._translate(c_irsb.next, self)
+            self.statements = [stmt.IRStmt._translate(c_irsb.stmts[i]) for i in xrange(c_irsb.stmts_used)]
+            self.next = expr.IRExpr._translate(c_irsb.next)
             self.offsIP = c_irsb.offsIP
             self.stmts_used = c_irsb.stmts_used
             self.jumpkind = ints_to_enums[c_irsb.jumpkind]
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -28,7 +28,7 @@ class IRSB(VEXObject):
                  ]
 
     def __init__(self, data, mem_addr, arch, num_inst=None, num_bytes=None, bytes_offset=0,
-                 traceflags=0):  # pylint:disable=redefined-builtin
+                 traceflags=0, empty=False):  # pylint:disable=redefined-builtin
         """
         :param data:            The bytes to lift. Can be either a string of bytes or a cffi buffer object.
         :type data:             str or bytes or cffi.FFI.CData
@@ -39,6 +39,7 @@ class IRSB(VEXObject):
         :param num_bytes:       The maximum number of bytes to use. Max 400.
         :param bytes_offset:    The offset into `data` to start lifting at.
         :param traceflags:      The libVEX traceflags, controlling VEX debug prints.
+        :param empty:           The empty flag, when True, returns an empty IRSB.
 
         .. note:: Explicitly specifying the number of instructions to lift (`num_inst`) may not always work
                   exactly as expected. For example, on MIPS, it is meaningless to lift a branch or jump
@@ -47,53 +48,29 @@ class IRSB(VEXObject):
                   on MIPS as a single instruction (`num_inst=1`) will result in an empty IRSB, and subsequent
                   attempts to run this block will raise `SimIRSBError('Empty IRSB passed to SimIRSB.')`.
         """
-        try:
-            _libvex_lock.acquire()
+        VEXObject.__init__(self)
 
-            VEXObject.__init__(self)
-
-            if isinstance(data, (str, bytes)):
-                num_bytes = len(data) if num_bytes is None else num_bytes
-                c_bytes = ffi.new('unsigned char [%d]' % (len(data) + 8), data + '\0' * 8)
-            else:
-                if not num_bytes:
-                    raise PyVEXError("C-backed bytes must have the length specified by num_bytes")
-                c_bytes = data
-
-            if num_bytes == 0:
-                raise PyVEXError("No bytes provided")
-            pvc.vta.traceflags = traceflags
-
-            vex_arch = getattr(pvc, arch.vex_arch)
-
-            arch.vex_archinfo['hwcache_info']['caches'] = ffi.NULL
-
-            if num_inst is not None:
-                c_irsb = pvc.vex_block_inst(vex_arch, arch.vex_archinfo, c_bytes + bytes_offset, mem_addr, num_inst)
-            else:
-                c_irsb = pvc.vex_block_bytes(vex_arch, arch.vex_archinfo, c_bytes + bytes_offset, mem_addr, num_bytes, 1)
-
-            if c_irsb == ffi.NULL:
-                raise PyVEXError(ffi.string(pvc.last_error) if pvc.last_error != ffi.NULL else "unknown error")
-
-            # We must use a pickle value, CData objects are not pickeable so not ffi.NULL
-            arch.vex_archinfo['hwcache_info']['caches'] = None
-
-            self.c_irsb = c_irsb
-            self.arch = arch
-            self.tyenv = IRTypeEnv._from_c(c_irsb.tyenv)
-            self.statements = [stmt.IRStmt._from_c(c_irsb.stmts[i]) for i in xrange(c_irsb.stmts_used)]
-            self.next = expr.IRExpr._from_c(c_irsb.next)
-            self.offsIP = c_irsb.offsIP
-            self.stmts_used = c_irsb.stmts_used
-            self.jumpkind = ints_to_enums[c_irsb.jumpkind]
-
+        if empty:
+            self.tyenv = None
+            self.statements = None
+            self.next = None
+            self.jumpkind = None
+            self.offsIP = None
             self._addr = mem_addr
-            self.direct_next = self._is_defaultexit_direct_jump()
+            self.arch = arch
+            self.direct_next = None
 
-            del self.c_irsb
-        finally:
-            _libvex_lock.release()
+            return
+        
+        irsb = IRSB.lift(data, mem_addr, arch, num_inst, num_bytes, bytes_offset, traceflags)
+        self.tyenv = irsb.tyenv
+        self.statements = irsb.statements
+        self.next = irsb.next
+        self.jumpkind = irsb.jumpkind
+        self.offsIP = irsb.offsIP
+        self._addr = irsb._addr
+        self.arch = irsb.arch
+        self.direct_next = self._is_defaultexit_direct_jump()
 
     def pp(self):
         """
@@ -260,6 +237,92 @@ class IRSB(VEXObject):
         target = self._get_defaultexit_target()
         return target is not None
 
+    @staticmethod
+    def lift(data, mem_addr, arch, num_inst=None, num_bytes=None, bytes_offset=0,
+             traceflags=0):  # pylint:disable=redefined-builtin
+        """
+        :param data:            The bytes to lift. Can be either a string of bytes or a cffi buffer object.
+        :type data:             str or bytes or cffi.FFI.CData
+        :param int mem_addr:    The address to lift the data at.
+        :param arch:            The architecture to lift the data as.
+        :type arch:             :class:`archinfo.Arch`
+        :param num_inst:        The maximum number of instructions to lift. Max 99. (See note below)
+        :param num_bytes:       The maximum number of bytes to use. Max 400.
+        :param bytes_offset:    The offset into `data` to start lifting at.
+        :param traceflags:      The libVEX traceflags, controlling VEX debug prints.
+
+        .. note:: Explicitly specifying the number of instructions to lift (`num_inst`) may not always work
+                  exactly as expected. For example, on MIPS, it is meaningless to lift a branch or jump
+                  instruction without its delay slot. VEX attempts to Do The Right Thing by possibly decoding
+                  fewer instructions than requested. Specifically, this means that lifting a branch or jump
+                  on MIPS as a single instruction (`num_inst=1`) will result in an empty IRSB, and subsequent
+                  attempts to run this block will raise `SimIRSBError('Empty IRSB passed to SimIRSB.')`.
+        """
+        irsb = None
+        
+        try:
+            _libvex_lock.acquire()
+
+            if isinstance(data, (str, bytes)):
+                num_bytes = len(data) if num_bytes is None else num_bytes
+                c_bytes = ffi.new('unsigned char [%d]' % (len(data) + 8), data + '\0' * 8)
+            else:
+                if not num_bytes:
+                    raise PyVEXError("C-backed bytes must have the length specified by num_bytes")
+                c_bytes = data
+
+            if num_bytes == 0:
+                raise PyVEXError("No bytes provided")
+            pvc.vta.traceflags = traceflags
+
+            vex_arch = getattr(pvc, arch.vex_arch)
+
+            arch.vex_archinfo['hwcache_info']['caches'] = ffi.NULL
+
+            if num_inst is not None:
+                c_irsb = pvc.vex_block_inst(vex_arch, arch.vex_archinfo, c_bytes + bytes_offset, mem_addr, num_inst)
+            else:
+                c_irsb = pvc.vex_block_bytes(vex_arch, arch.vex_archinfo, c_bytes + bytes_offset, mem_addr, num_bytes, 1)
+
+            if c_irsb == ffi.NULL:
+                raise PyVEXError(ffi.string(pvc.last_error) if pvc.last_error != ffi.NULL else "unknown error")
+
+            # We must use a pickle value, CData objects are not pickeable so not ffi.NULL
+            arch.vex_archinfo['hwcache_info']['caches'] = None
+            
+            irsb = IRSB._from_c(c_irsb, mem_addr, arch)
+            
+            del c_irsb
+        finally:
+            _libvex_lock.release()
+            
+        return irsb
+            
+    @staticmethod
+    def _from_c(c_irsb, mem_addr, arch):
+        stmts = [stmt.IRStmt._from_c(c_irsb.stmts[i]) for i in xrange(c_irsb.stmts_used)]
+        return IRSB._from_py(IRTypeEnv._from_c(c_irsb.tyenv),
+                             stmts,
+                             c_irsb.stmts_used,
+                             expr.IRExpr._from_c(c_irsb.next),
+                             ints_to_enums[c_irsb.jumpkind],
+                             c_irsb.offsIP,
+                             mem_addr,
+                             arch)
+
+    @staticmethod
+    def _from_py(tyenv, stmts, stmts_used, next, jumpkind, offsIP, mem_addr, arch):
+        irsb = IRSB(None, mem_addr, arch, empty=True)
+
+        irsb.tyenv = tyenv
+        irsb.statements = stmts
+        irsb.stmts_used = stmts_used
+        irsb.next = next
+        irsb.jumpkind = jumpkind
+        irsb.offsIP = offsIP
+        irsb.direct_next = irsb._is_defaultexit_direct_jump()
+
+        return irsb
 
 class IRTypeEnv(VEXObject):
     """
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -112,7 +112,14 @@ class IRSB(VEXObject):
         sa.append("   %s" % self.tyenv)
         sa.append("")
         for i, s in enumerate(self.statements):
-            sa.append("   %02d | %s" % (i, s))
+            stmt_str = ''
+            if isinstance(s, stmt.Put):
+                stmt_str = s.__str__(reg_name=self.arch.translate_register_name(s.offset, s.data.result_size(self.tyenv)/8))
+            elif isinstance(s, stmt.WrTmp) and isinstance(s.data, expr.Get):
+                stmt_str = s.__str__(reg_name=self.arch.translate_register_name(s.data.offset, s.data.result_size(self.tyenv)/8))
+            else:
+                stmt_str = s.__str__()
+            sa.append("   %02d | %s" % (i, stmt_str))
         sa.append(
             "   NEXT: PUT(%s) = %s; %s" % (self.arch.translate_register_name(self.offsIP), self.next, self.jumpkind))
         sa.append("}")
--- a/pyvex/const.py
+++ b/pyvex/const.py
@@ -6,10 +6,10 @@ class IRConst(VEXObject):
     __slots__ = ['tag', 'value']
 
     type = None
+    tag = None
 
-    def __init__(self, c_expr):
+    def __init__(self):
         VEXObject.__init__(self)
-        self.tag = ints_to_enums[c_expr.tag]
 
     def pp(self):
         print self.__str__()
@@ -23,122 +23,177 @@ class IRConst(VEXObject):
         if c_expr[0] == ffi.NULL:
             return None
 
-        tag = c_expr.tag
+        tag_int = c_expr.tag
 
         try:
-            return tag_to_class[tag](c_expr)
+            return tag_to_class[tag_int].from_c(c_expr)
         except KeyError:
-            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag])
+            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag_int])
 
 class U1(IRConst):
     type = 'Ity_I1'
-
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.U1
+    tag = 'Ico_U1'
+    
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "%d" % self.value
 
+    @staticmethod
+    def from_c(c_expr):
+        return U1(c_expr.Ico.U1)
+
 class U8(IRConst):
     type = 'Ity_I8'
-
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.U8
+    tag = 'Ico_U8'
+    
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "0x%02x" % self.value
 
+    @staticmethod
+    def from_c(c_expr):
+        return U8(c_expr.Ico.U8)
+
 class U16(IRConst):
     type = 'Ity_I16'
+    tag = 'Ico_U16'
 
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.U16
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "0x%04x" % self.value
 
+    @staticmethod
+    def from_c(c_expr):
+        return U16(c_expr.Ico.U16)
+
 class U32(IRConst):
     type = 'Ity_I32'
-
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.U32
+    tag = 'Ico_U32'
+    
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "0x%08x" % self.value
 
+    @staticmethod
+    def from_c(c_expr):
+        return U32(c_expr.Ico.U32)
+
 class U64(IRConst):
     type = 'Ity_I64'
+    tag = 'Ico_U64'
 
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.U64
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "0x%016x" % self.value
 
+    @staticmethod
+    def from_c(c_expr):
+        return U64(c_expr.Ico.U64)
+
 class F32(IRConst):
     type = 'Ity_F32'
-
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.F32
+    tag = 'Ico_F32'
+    
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "%f" % self.value
+
+    @staticmethod
+    def from_c(c_expr):
+        return F32(c_expr.Ico.F32)
 
 class F32i(IRConst):
     type = 'Ity_F32'
+    tag = 'Ico_F32i'
 
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.F32
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "%f" % self.value
+
+    @staticmethod
+    def from_c(c_expr):
+        return F32i(c_expr.Ico.F32)
 
 class F64(IRConst):
     type = 'Ity_F64'
+    tag = 'Ico_F64'
 
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.F64
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "%f" % self.value
+
+    @staticmethod
+    def from_c(c_expr):
+        return F64(c_expr.Ico.F64)
 
 class F64i(IRConst):
     type = 'Ity_F64'
+    tag = 'Ico_F64i'
 
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.F64
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "%f" % self.value
 
+    @staticmethod
+    def from_c(c_expr):
+        return F64i(c_expr.Ico.F64)
+
 class V128(IRConst):
     type = 'Ity_V128'
-
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.V128
+    tag = 'Ico_V128'
+    
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "%x" % self.value
+
+    @staticmethod
+    def from_c(c_expr):
+        return V128(c_expr.Ico.V128)
 
 class V256(IRConst):
     type = 'Ity_V256'
+    tag = 'Ico_V256'
 
-    def __init__(self, c_expr):
-        IRConst.__init__(self, c_expr)
-        self.value = c_expr.Ico.V256
+    def __init__(self, value):
+        IRConst.__init__(self)
+        self.value = value
 
     def __str__(self):
         return "%x" % self.value
+
+    @staticmethod
+    def from_c(c_expr):
+        return V256(c_expr.Ico.V256)
 
 from .enums import ints_to_enums, enums_to_ints, type_sizes
 from .errors import PyVEXError
--- a/pyvex/const.py
+++ b/pyvex/const.py
@@ -19,16 +19,29 @@ class IRConst(VEXObject):
         return type_sizes[self.type]
 
     @staticmethod
-    def _translate(c_expr):
-        if c_expr[0] == ffi.NULL:
+    def _from_c(c_const):
+        if c_const[0] == ffi.NULL:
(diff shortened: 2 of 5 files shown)


mattrepl:
Just realized I haven't updated the `IRSB` class be constructed from Python objects. Making that change now and will update the PR.

rhelmot:
When you make that change, please do it with a static method, don't touch the IRSB class constructor. That interface is used by pretty much everyone, and breaking it Will cause problems.

mattrepl:
Ok, added `lift`, `_from_c` and, and `_from_py` static methods to the `IRSB` class. The interface to `__init__` remains the same, but under the hood it uses `lift`. If someone wants to create an IRSB instance from pure Python and without lifting from bytes, they'll need to use `_from_py`.

mattrepl:
The `CCall` class in pyvex/expr.py was missing a `_to_c` method. It's no longer missing.

rhelmot:
There appear to be several issues with this. Running tests nets me numerous failures immediately. The only one I've triaged so far is because ITE is missing a `_to_c` method, which I've fixed.

Unfortunately, the structure of github is gonna prevent us from working collaboratively on bugfixes since there's no one repository that we both have push access to. I think at this point the best thing to do is for me to post the bugs I've fixed so far as diffs, and for you to post the bugs you've fixed so far as diffs.

To run tests for yourself, you can go to the `angr/tests` folder and run `nosetests -v --nocapture --nologcapture test_*.py`

mattrepl:
Interesting, I was having unrelated issues running the pyvex tests. It looked like they had suffered from bit rot and I had even started a local branch to update them. I'll check again.

mattrepl:
Nevermind, just saw you referenced the angr tests, not pyvex.

rhelmot:
The pyvex tests are 100% broken! They are not used at all, and are designed to test a version of pyvex several generations old. Use the angr tests.

rhelmot:
```diff
diff --git a/pyvex/expr.py b/pyvex/expr.py
index 1ffdf7b..8fa8bf4 100644
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -457,6 +457,12 @@ class ITE(IRExpr):
                    IRExpr._from_c(c_expr.Iex.ITE.iffalse),
                    IRExpr._from_c(c_expr.Iex.ITE.iftrue))

+    @staticmethod
+    def _to_c(expr):
+        return pvc.IRExpr_ITE(ITE._to_c(expr.cond),
+                              ITE._to_c(expr.iftrue),
+                              ITE._to_c(expr.iffalse))
+
 class CCall(IRExpr):
     """
     A call to a pure (no side-effects) helper C function.
```

rhelmot:
Oh hell, that patch was wrong. Here's the right version.

```diff
diff --git a/pyvex/expr.py b/pyvex/expr.py
index 1ffdf7b..cf2df29 100644
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -457,6 +457,12 @@ class ITE(IRExpr):
                    IRExpr._from_c(c_expr.Iex.ITE.iffalse),
                    IRExpr._from_c(c_expr.Iex.ITE.iftrue))

+    @staticmethod
+    def _to_c(expr):
+        return pvc.IRExpr_ITE(IRExpr._to_c(expr.cond),
+                              IRExpr._to_c(expr.iftrue),
+                              IRExpr._to_c(expr.iffalse))
+
 class CCall(IRExpr):
     """
     A call to a pure (no side-effects) helper C function.
```

mattrepl:
Thank! Will suss out any other bugs and update once all relevant angr tests seem to pass.

rhelmot:
```diff
diff --git a/pyvex/stmt.py b/pyvex/stmt.py
index c5b1b3b..1ea9d45 100644
--- a/pyvex/stmt.py
+++ b/pyvex/stmt.py
@@ -343,7 +343,7 @@ class Dirty(IRStmt):

     @staticmethod
     def _from_c(c_stmt):
-        return Dirty(IRCallee(c_stmt.Ist.Dirty.details.cee),
+        return Dirty(IRCallee._from_c(c_stmt.Ist.Dirty.details.cee),
                      IRExpr._from_c(c_stmt.Ist.Dirty.details.guard),
                      tuple([IRExpr._from_c(arg)
                             for arg in itertools.takewhile(lambda a: a != ffi.NULL,
```

rhelmot:
```diff
diff --git a/pyvex/stmt.py b/pyvex/stmt.py
index 1ea9d45..e2d95b6 100644
--- a/pyvex/stmt.py
+++ b/pyvex/stmt.py
@@ -343,11 +343,17 @@ class Dirty(IRStmt):

     @staticmethod
     def _from_c(c_stmt):
+        args = []
+        for i in xrange(20):
+            a = c_stmt.Ist.Dirty.details.args[i]
+            if a == ffi.NULL:
+                break
+
+            args.append(IRExpr._from_c(a))
+
         return Dirty(IRCallee._from_c(c_stmt.Ist.Dirty.details.cee),
                      IRExpr._from_c(c_stmt.Ist.Dirty.details.guard),
-                     tuple([IRExpr._from_c(arg)
-                            for arg in itertools.takewhile(lambda a: a != ffi.NULL,
-                                                           c_stmt.Ist.Dirty.details.args)]),
+                     tuple(args),
                      c_stmt.Ist.Dirty.details.tmp,
                      ints_to_enums[c_stmt.Ist.Dirty.details.mFx],
                      IRExpr._from_c(c_stmt.Ist.Dirty.details.mAddr),
```

rhelmot:
ok, um, I'm now running into the issue where libvex itself is throwing a "ran out of temporary storage" error and it looks like it's because you're using `_to_c` extremely judiciously for the `result_type`  calls without cleaning up after yourself.

My opinion here is that calling into libvex for anything other than lifting is a really really really really really really bad idea. I'm going to try to clean this up, but I have no idea how messy it'll be.

rhelmot:
Yeah, I'm going to restore the old system of each expression calculating its own `result_type` field. Sorry.

rhelmot:
...sorry, that made it sound like I was going to revert the simuvex PR. I just mean I'm going to move the definition of `result_type()` onto each individual expression class, which will return a value without calling into libvex.

mattrepl:
Hmm, ok. What if we just made sure to cleanup the references after? I don't disagree about making calls into libvex rare, but there may end up being a fair amount of code to rewrite in Python.

mattrepl:
Lemme take a look now that I know there are tests that will exercise bugs and runtime issues I didn't consider.

rhelmot:
This implements what I described. the only complex code that would have had to be copied was the code to map IROps to types, so I called into a simpler libvex function that has no side effects, typeOfPrimop. This function was not previously exported, so you'll need to rebuild our fork of libvex.

```diff
diff --git a/pyvex/block.py b/pyvex/block.py
index 26735e5..47c4386 100644
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -343,6 +343,14 @@ class IRTypeEnv(VEXObject):
     def __str__(self):
         return ' '.join(("t%d:%s" % (i, t)) for i, t in enumerate(self.types))
 
+    def lookup(self, tmp):
+        """
+        Return the type of temporary variable `tmp` as an enum string
+        """
+        if tmp < 0 or tmp > self.types_used:
+            raise IndexError(tmp)
+        return self.types[tmp]
+
     @staticmethod
     def _from_c(c_tyenv):
         return IRTypeEnv([ints_to_enums[c_tyenv.types[t]] for t in xrange(c_tyenv.types_used)],
diff --git a/pyvex/expr.py b/pyvex/expr.py
index cf2df29..680be83 100644
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -8,10 +8,10 @@ class IRExpr(VEXObject):
     __slots__ = ['tag']
 
     tag = None
-    
+
     def __init__(self):
         VEXObject.__init__(self)
-    
+
     def pp(self):
         print self.__str__()
 
@@ -46,10 +46,7 @@ class IRExpr(VEXObject):
         return type_sizes[self.result_type(tyenv)]
 
     def result_type(self, tyenv):
-        if isinstance(self, (VECRET, Binder, BBPTR)):
-            return 'Ity_INVALID'
-        else:
-            return ints_to_enums[pvc.typeOfIRExpr(IRTypeEnv._to_c(tyenv), IRExpr._to_c(self))]
+        raise NotImplementedError
 
     @staticmethod
     def _from_c(c_expr):
@@ -61,7 +58,7 @@ class IRExpr(VEXObject):
         try:
             return tag_to_class[tag_int]._from_c(c_expr)
         except KeyError:
-            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag_int])            
+            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag_int])
     _translate = _from_c
 
     @staticmethod
@@ -84,7 +81,7 @@ class Binder(IRExpr):
     def __init__(self, binder):
         IRExpr.__init__(self)
         self.binder = binder
-    
+
     def __str__(self):
         return "Binder"
 
@@ -96,10 +93,13 @@ class Binder(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_Binder(expr.binder)
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
 class VECRET(IRExpr):
 
     tag = 'Iex_VECRET'
-    
+
     def __init__(self):
         IRExpr.__init__(self)
 
@@ -114,10 +114,14 @@ class VECRET(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_VECRET()
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
+
 class BBPTR(IRExpr):
 
     tag = 'Iex_BBPTR'
-    
+
     def __init__(self):
         IRExpr.__init__(self)
 
@@ -132,6 +136,10 @@ class BBPTR(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_BBPTR()
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
+
 class GetI(IRExpr):
     """
     Read a guest register at a non-fixed offset in the guest state.
@@ -146,7 +154,7 @@ class GetI(IRExpr):
         self.descr = descr
         self.ix = ix
         self.bias = bias
-    
+
     @property
     def description(self):
         return self.descr
@@ -171,6 +179,10 @@ class GetI(IRExpr):
                                IRExpr._to_c(expr.ix),
                                expr.bias)
 
+    def result_type(self, tyenv):
+        return self.descr.elemTy
+
+
 class RdTmp(IRExpr):
     """
     Read the value held by a temporary.
@@ -195,6 +207,10 @@ class RdTmp(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_RdTmp(expr.tmp)
 
+    def result_type(self, tyenv):
+        return tyenv.lookup(self.tmp)
+
+
 class Get(IRExpr):
     """
     Read a guest register, at a fixed offset in the guest state.
@@ -203,7 +219,7 @@ class Get(IRExpr):
     __slots__ = ['offset', 'ty']
 
     tag = 'Iex_Get'
-    
+
     def __init__(self, offset, ty):
         IRExpr.__init__(self)
         self.offset = offset
@@ -212,7 +228,7 @@ class Get(IRExpr):
     @property
     def type(self):
         return self.ty
-    
+
     def __str__(self, reg_name=None):
         if reg_name:
             return "GET:%s(%s)" % (self.ty[4:], reg_name)
@@ -228,7 +244,11 @@ class Get(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_Get(expr.offset,
                               enums_to_ints[expr.ty])
-        
+
+    def result_type(self, tyenv):
+        return self.ty
+
+
 class Qop(IRExpr):
     """
     A quaternary operation (4 arguments).
@@ -242,7 +262,7 @@ class Qop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -267,6 +287,9 @@ class Qop(IRExpr):
                               *[IRExpr._to_c(arg)
                                 for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
 class Triop(IRExpr):
     """
     A ternary operation (3 arguments)
@@ -280,7 +303,7 @@ class Triop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -304,6 +327,10 @@ class Triop(IRExpr):
                                 *[IRExpr._to_c(arg)
                                   for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Binop(IRExpr):
     """
     A binary operation (2 arguments).
@@ -317,7 +344,7 @@ class Binop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -340,6 +367,10 @@ class Binop(IRExpr):
                                 *[IRExpr._to_c(arg)
                                   for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Unop(IRExpr):
     """
     A unary operation (1 argument).
@@ -350,9 +381,10 @@ class Unop(IRExpr):
     tag = 'Iex_Unop'
 
     def __init__(self, op, args):
+        IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -372,6 +404,10 @@ class Unop(IRExpr):
         return pvc.IRExpr_Unop(enums_to_ints[expr.op],
                                IRExpr._to_c(expr.args[0]))
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Load(IRExpr):
     """
     A load from memory.
@@ -386,7 +422,7 @@ class Load(IRExpr):
         self.end = end
         self.ty = ty
         self.addr = addr
-    
+
     @property
     def endness(self):
         return self.end
@@ -410,6 +446,10 @@ class Load(IRExpr):
                                enums_to_ints[expr.ty],
                                IRExpr._to_c(expr.addr))
 
+    def result_type(self, tyenv):
+        return self.ty
+
+
 class Const(IRExpr):
     """
     A constant expression.
@@ -420,8 +460,9 @@ class Const(IRExpr):
     tag = 'Iex_Const'
 
     def __init__(self, con):
+        IRExpr.__init__(self)
         self.con = con
-    
+
     def __str__(self):
         return str(self.con)
 
@@ -432,7 +473,11 @@ class Const(IRExpr):
     @staticmethod
     def _to_c(expr):
         return pvc.IRExpr_Const(IRConst._to_c(expr.con))
-        
+
+    def result_type(self, tyenv):
+        return self.con.type
+
+
 class ITE(IRExpr):
     """
     An if-then-else expression.
@@ -463,6 +508,9 @@ class ITE(IRExpr):
                               IRExpr._to_c(expr.iftrue),
                               IRExpr._to_c(expr.iffalse))
 
+    def result_type(self, tyenv):
+        return self.iftrue.result_type(tyenv)
+
 class CCall(IRExpr):
     """
     A call to a pure (no side-effects) helper C function.
@@ -477,7 +525,7 @@ class CCall(IRExpr):
         self.retty = retty
         self.cee = cee
         self.args = tuple(args)
-    
+
     @property
     def ret_type(self):
         return self.retty
@@ -505,7 +553,7 @@ class CCall(IRExpr):
                 break
             args.append(IRExpr._from_c(arg))
             i += 1
-            
+
         return CCall(ints_to_enums[c_expr.Iex.CCall.retty],
                      IRCallee._from_c(c_expr.Iex.CCall.cee),
                      tuple(args))
@@ -517,7 +565,22 @@ class CCall(IRExpr):
                                 enums_to_ints[expr.retty],
                                 mkIRExprVec[len(args)](*args))
 
-from .block import IRTypeEnv
+    def result_type(self, tyenv):
+        return self.retty
+
+_op_type_cache = {}
+
+def op_type(op):
+    try:
+        return _op_type_cache[op]
+    except KeyError:
+        out_int = ffi.new('IRType *')
+        unused = ffi.new('IRType *')
+        pvc.typeOfPrimop(enums_to_ints[op], out_int, unused, unused, unused, unused)
+        out = ints_to_enums[out_int[0]]
+        _op_type_cache[op] = out
+        return out
+
 from .const import IRConst
 from .enums import IRCallee, IRRegArray, enums_to_ints, ints_to_enums, type_sizes
 from .errors import PyVEXErrordiff --git a/pyvex/block.py b/pyvex/block.py
index 26735e5..47c4386 100644
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -343,6 +343,14 @@ class IRTypeEnv(VEXObject):
     def __str__(self):
         return ' '.join(("t%d:%s" % (i, t)) for i, t in enumerate(self.types))
 
+    def lookup(self, tmp):
+        """
+        Return the type of temporary variable `tmp` as an enum string
+        """
+        if tmp < 0 or tmp > self.types_used:
+            raise IndexError(tmp)
+        return self.types[tmp]
+
     @staticmethod
     def _from_c(c_tyenv):
         return IRTypeEnv([ints_to_enums[c_tyenv.types[t]] for t in xrange(c_tyenv.types_used)],
diff --git a/pyvex/expr.py b/pyvex/expr.py
index cf2df29..680be83 100644
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -8,10 +8,10 @@ class IRExpr(VEXObject):
     __slots__ = ['tag']
 
     tag = None
-    
+
     def __init__(self):
         VEXObject.__init__(self)
-    
+
     def pp(self):
         print self.__str__()
 
@@ -46,10 +46,7 @@ class IRExpr(VEXObject):
         return type_sizes[self.result_type(tyenv)]
 
     def result_type(self, tyenv):
-        if isinstance(self, (VECRET, Binder, BBPTR)):
-            return 'Ity_INVALID'
-        else:
-            return ints_to_enums[pvc.typeOfIRExpr(IRTypeEnv._to_c(tyenv), IRExpr._to_c(self))]
+        raise NotImplementedError
 
     @staticmethod
     def _from_c(c_expr):
@@ -61,7 +58,7 @@ class IRExpr(VEXObject):
         try:
             return tag_to_class[tag_int]._from_c(c_expr)
         except KeyError:
-            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag_int])            
+            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag_int])
     _translate = _from_c
 
     @staticmethod
@@ -84,7 +81,7 @@ class Binder(IRExpr):
     def __init__(self, binder):
         IRExpr.__init__(self)
         self.binder = binder
-    
+
     def __str__(self):
         return "Binder"
 
@@ -96,10 +93,13 @@ class Binder(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_Binder(expr.binder)
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
 class VECRET(IRExpr):
 
     tag = 'Iex_VECRET'
-    
+
     def __init__(self):
         IRExpr.__init__(self)
 
@@ -114,10 +114,14 @@ class VECRET(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_VECRET()
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
+
 class BBPTR(IRExpr):
 
     tag = 'Iex_BBPTR'
-    
+
     def __init__(self):
         IRExpr.__init__(self)
 
@@ -132,6 +136,10 @@ class BBPTR(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_BBPTR()
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
+
 class GetI(IRExpr):
     """
     Read a guest register at a non-fixed offset in the guest state.
@@ -146,7 +154,7 @@ class GetI(IRExpr):
         self.descr = descr
         self.ix = ix
         self.bias = bias
-    
+
     @property
     def description(self):
         return self.descr
@@ -171,6 +179,10 @@ class GetI(IRExpr):
                                IRExpr._to_c(expr.ix),
                                expr.bias)
 
+    def result_type(self, tyenv):
+        return self.descr.elemTy
+
+
 class RdTmp(IRExpr):
     """
     Read the value held by a temporary.
@@ -195,6 +207,10 @@ class RdTmp(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_RdTmp(expr.tmp)
 
+    def result_type(self, tyenv):
+        return tyenv.lookup(self.tmp)
+
+
 class Get(IRExpr):
     """
     Read a guest register, at a fixed offset in the guest state.
@@ -203,7 +219,7 @@ class Get(IRExpr):
     __slots__ = ['offset', 'ty']
 
     tag = 'Iex_Get'
-    
+
     def __init__(self, offset, ty):
         IRExpr.__init__(self)
         self.offset = offset
@@ -212,7 +228,7 @@ class Get(IRExpr):
     @property
     def type(self):
         return self.ty
-    
+
     def __str__(self, reg_name=None):
         if reg_name:
             return "GET:%s(%s)" % (self.ty[4:], reg_name)
@@ -228,7 +244,11 @@ class Get(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_Get(expr.offset,
                               enums_to_ints[expr.ty])
-        
+
+    def result_type(self, tyenv):
+        return self.ty
+
+
 class Qop(IRExpr):
     """
     A quaternary operation (4 arguments).
@@ -242,7 +262,7 @@ class Qop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -267,6 +287,9 @@ class Qop(IRExpr):
                               *[IRExpr._to_c(arg)
                                 for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
 class Triop(IRExpr):
     """
     A ternary operation (3 arguments)
@@ -280,7 +303,7 @@ class Triop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -304,6 +327,10 @@ class Triop(IRExpr):
                                 *[IRExpr._to_c(arg)
                                   for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Binop(IRExpr):
     """
     A binary operation (2 arguments).
@@ -317,7 +344,7 @@ class Binop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -340,6 +367,10 @@ class Binop(IRExpr):
                                 *[IRExpr._to_c(arg)
                                   for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Unop(IRExpr):
     """
     A unary operation (1 argument).
@@ -350,9 +381,10 @@ class Unop(IRExpr):
     tag = 'Iex_Unop'
 
     def __init__(self, op, args):
+        IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -372,6 +404,10 @@ class Unop(IRExpr):
         return pvc.IRExpr_Unop(enums_to_ints[expr.op],
                                IRExpr._to_c(expr.args[0]))
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Load(IRExpr):
     """
     A load from memory.
@@ -386,7 +422,7 @@ class Load(IRExpr):
         self.end = end
         self.ty = ty
         self.addr = addr
-    
+
     @property
     def endness(self):
         return self.end
@@ -410,6 +446,10 @@ class Load(IRExpr):
                                enums_to_ints[expr.ty],
                                IRExpr._to_c(expr.addr))
 
+    def result_type(self, tyenv):
+        return self.ty
+
+
 class Const(IRExpr):
     """
     A constant expression.
@@ -420,8 +460,9 @@ class Const(IRExpr):
     tag = 'Iex_Const'
 
     def __init__(self, con):
+        IRExpr.__init__(self)
         self.con = con
-    
+
     def __str__(self):
         return str(self.con)
 
@@ -432,7 +473,11 @@ class Const(IRExpr):
     @staticmethod
     def _to_c(expr):
         return pvc.IRExpr_Const(IRConst._to_c(expr.con))
-        
+
+    def result_type(self, tyenv):
+        return self.con.type
+
+
 class ITE(IRExpr):
     """
     An if-then-else expression.
@@ -463,6 +508,9 @@ class ITE(IRExpr):
                               IRExpr._to_c(expr.iftrue),
                               IRExpr._to_c(expr.iffalse))
 
+    def result_type(self, tyenv):
+        return self.iftrue.result_type(tyenv)
+
 class CCall(IRExpr):
     """
     A call to a pure (no side-effects) helper C function.
@@ -477,7 +525,7 @@ class CCall(IRExpr):
         self.retty = retty
         self.cee = cee
         self.args = tuple(args)
-    
+
     @property
     def ret_type(self):
         return self.retty
@@ -505,7 +553,7 @@ class CCall(IRExpr):
                 break
             args.append(IRExpr._from_c(arg))
             i += 1
-            
+
         return CCall(ints_to_enums[c_expr.Iex.CCall.retty],
                      IRCallee._from_c(c_expr.Iex.CCall.cee),
                      tuple(args))
@@ -517,7 +565,22 @@ class CCall(IRExpr):
                                 enums_to_ints[expr.retty],
                                 mkIRExprVec[len(args)](*args))
 
-from .block import IRTypeEnv
+    def result_type(self, tyenv):
+        return self.retty
+
+_op_type_cache = {}
+
+def op_type(op):
+    try:
+        return _op_type_cache[op]
+    except KeyError:
+        out_int = ffi.new('IRType *')
+        unused = ffi.new('IRType *')
+        pvc.typeOfPrimop(enums_to_ints[op], out_int, unused, unused, unused, unused)
+        out = ints_to_enums[out_int[0]]
+        _op_type_cache[op] = out
+        return out
+
 from .const import IRConst
 from .enums import IRCallee, IRRegArray, enums_to_ints, ints_to_enums, type_sizes
 from .errors import PyVEXErrordiff --git a/pyvex/block.py b/pyvex/block.py
index 26735e5..47c4386 100644
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -343,6 +343,14 @@ class IRTypeEnv(VEXObject):
     def __str__(self):
         return ' '.join(("t%d:%s" % (i, t)) for i, t in enumerate(self.types))
 
+    def lookup(self, tmp):
+        """
+        Return the type of temporary variable `tmp` as an enum string
+        """
+        if tmp < 0 or tmp > self.types_used:
+            raise IndexError(tmp)
+        return self.types[tmp]
+
     @staticmethod
     def _from_c(c_tyenv):
         return IRTypeEnv([ints_to_enums[c_tyenv.types[t]] for t in xrange(c_tyenv.types_used)],
diff --git a/pyvex/expr.py b/pyvex/expr.py
index cf2df29..680be83 100644
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -8,10 +8,10 @@ class IRExpr(VEXObject):
     __slots__ = ['tag']
 
     tag = None
-    
+
     def __init__(self):
         VEXObject.__init__(self)
-    
+
     def pp(self):
         print self.__str__()
 
@@ -46,10 +46,7 @@ class IRExpr(VEXObject):
         return type_sizes[self.result_type(tyenv)]
 
     def result_type(self, tyenv):
-        if isinstance(self, (VECRET, Binder, BBPTR)):
-            return 'Ity_INVALID'
-        else:
-            return ints_to_enums[pvc.typeOfIRExpr(IRTypeEnv._to_c(tyenv), IRExpr._to_c(self))]
+        raise NotImplementedError
 
     @staticmethod
     def _from_c(c_expr):
@@ -61,7 +58,7 @@ class IRExpr(VEXObject):
         try:
             return tag_to_class[tag_int]._from_c(c_expr)
         except KeyError:
-            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag_int])            
+            raise PyVEXError('Unknown/unsupported IRExprTag %s\n' % ints_to_enums[tag_int])
     _translate = _from_c
 
     @staticmethod
@@ -84,7 +81,7 @@ class Binder(IRExpr):
     def __init__(self, binder):
         IRExpr.__init__(self)
         self.binder = binder
-    
+
     def __str__(self):
         return "Binder"
 
@@ -96,10 +93,13 @@ class Binder(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_Binder(expr.binder)
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
 class VECRET(IRExpr):
 
     tag = 'Iex_VECRET'
-    
+
     def __init__(self):
         IRExpr.__init__(self)
 
@@ -114,10 +114,14 @@ class VECRET(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_VECRET()
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
+
 class BBPTR(IRExpr):
 
     tag = 'Iex_BBPTR'
-    
+
     def __init__(self):
         IRExpr.__init__(self)
 
@@ -132,6 +136,10 @@ class BBPTR(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_BBPTR()
 
+    def result_type(self, tyenv):
+        return 'Ity_INVALID'
+
+
 class GetI(IRExpr):
     """
     Read a guest register at a non-fixed offset in the guest state.
@@ -146,7 +154,7 @@ class GetI(IRExpr):
         self.descr = descr
         self.ix = ix
         self.bias = bias
-    
+
     @property
     def description(self):
         return self.descr
@@ -171,6 +179,10 @@ class GetI(IRExpr):
                                IRExpr._to_c(expr.ix),
                                expr.bias)
 
+    def result_type(self, tyenv):
+        return self.descr.elemTy
+
+
 class RdTmp(IRExpr):
     """
     Read the value held by a temporary.
@@ -195,6 +207,10 @@ class RdTmp(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_RdTmp(expr.tmp)
 
+    def result_type(self, tyenv):
+        return tyenv.lookup(self.tmp)
+
+
 class Get(IRExpr):
     """
     Read a guest register, at a fixed offset in the guest state.
@@ -203,7 +219,7 @@ class Get(IRExpr):
     __slots__ = ['offset', 'ty']
 
     tag = 'Iex_Get'
-    
+
     def __init__(self, offset, ty):
         IRExpr.__init__(self)
         self.offset = offset
@@ -212,7 +228,7 @@ class Get(IRExpr):
     @property
     def type(self):
         return self.ty
-    
+
     def __str__(self, reg_name=None):
         if reg_name:
             return "GET:%s(%s)" % (self.ty[4:], reg_name)
@@ -228,7 +244,11 @@ class Get(IRExpr):
     def _to_c(expr):
         return pvc.IRExpr_Get(expr.offset,
                               enums_to_ints[expr.ty])
-        
+
+    def result_type(self, tyenv):
+        return self.ty
+
+
 class Qop(IRExpr):
     """
     A quaternary operation (4 arguments).
@@ -242,7 +262,7 @@ class Qop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -267,6 +287,9 @@ class Qop(IRExpr):
                               *[IRExpr._to_c(arg)
                                 for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
 class Triop(IRExpr):
     """
     A ternary operation (3 arguments)
@@ -280,7 +303,7 @@ class Triop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -304,6 +327,10 @@ class Triop(IRExpr):
                                 *[IRExpr._to_c(arg)
                                   for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Binop(IRExpr):
     """
     A binary operation (2 arguments).
@@ -317,7 +344,7 @@ class Binop(IRExpr):
         IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -340,6 +367,10 @@ class Binop(IRExpr):
                                 *[IRExpr._to_c(arg)
                                   for arg in expr.args])
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Unop(IRExpr):
     """
     A unary operation (1 argument).
@@ -350,9 +381,10 @@ class Unop(IRExpr):
     tag = 'Iex_Unop'
 
     def __init__(self, op, args):
+        IRExpr.__init__(self)
         self.op = op
         self.args = args
-    
+
     def __str__(self):
         return "%s(%s)" % (self.op[4:], ','.join(str(a) for a in self.args))
 
@@ -372,6 +404,10 @@ class Unop(IRExpr):
         return pvc.IRExpr_Unop(enums_to_ints[expr.op],
                                IRExpr._to_c(expr.args[0]))
 
+    def result_type(self, tyenv):
+        return op_type(self.op)
+
+
 class Load(IRExpr):
     """
     A load from memory.
@@ -386,7 +422,7 @@ class Load(IRExpr):
         self.end = end
         self.ty = ty
         self.addr = addr
-    
+
     @property
     def endness(self):
         return self.end
@@ -410,6 +446,10 @@ class Load(IRExpr):
                                enums_to_ints[expr.ty],
                                IRExpr._to_c(expr.addr))
 
+    def result_type(self, tyenv):
+        return self.ty
+
+
 class Const(IRExpr):
     """
     A constant expression.
@@ -420,8 +460,9 @@ class Const(IRExpr):
     tag = 'Iex_Const'
 
     def __init__(self, con):
+        IRExpr.__init__(self)
         self.con = con
-    
+
     def __str__(self):
         return str(self.con)
 
@@ -432,7 +473,11 @@ class Const(IRExpr):
     @staticmethod
     def _to_c(expr):
         return pvc.IRExpr_Const(IRConst._to_c(expr.con))
-        
+
+    def result_type(self, tyenv):
+        return self.con.type
+
+
 class ITE(IRExpr):
     """
     An if-then-else expression.
@@ -463,6 +508,9 @@ class ITE(IRExpr):
                               IRExpr._to_c(expr.iftrue),
                               IRExpr._to_c(expr.iffalse))
 
+    def result_type(self, tyenv):
+        return self.iftrue.result_type(tyenv)
+
 class CCall(IRExpr):
     """
     A call to a pure (no side-effects) helper C function.
@@ -477,7 +525,7 @@ class CCall(IRExpr):
         self.retty = retty
         self.cee = cee
         self.args = tuple(args)
-    
+
     @property
     def ret_type(self):
         return self.retty
@@ -505,7 +553,7 @@ class CCall(IRExpr):
                 break
             args.append(IRExpr._from_c(arg))
             i += 1
-            
+
         return CCall(ints_to_enums[c_expr.Iex.CCall.retty],
                      IRCallee._from_c(c_expr.Iex.CCall.cee),
                      tuple(args))
@@ -517,7 +565,22 @@ class CCall(IRExpr):
                                 enums_to_ints[expr.retty],
                                 mkIRExprVec[len(args)](*args))
 
-from .block import IRTypeEnv
+    def result_type(self, tyenv):
+        return self.retty
+
+_op_type_cache = {}
+
+def op_type(op):
+    try:
+        return _op_type_cache[op]
+    except KeyError:
+        out_int = ffi.new('IRType *')
+        unused = ffi.new('IRType *')
+        pvc.typeOfPrimop(enums_to_ints[op], out_int, unused, unused, unused, unused)
+        out = ints_to_enums[out_int[0]]
+        _op_type_cache[op] = out
+        return out
+
 from .const import IRConst
 from .enums import IRCallee, IRRegArray, enums_to_ints, ints_to_enums, type_sizes
 from .errors import PyVEXError
```

rhelmot:
```diff
diff --git a/pyvex/const.py b/pyvex/const.py
index e86d59e..ddb833c 100644
--- a/pyvex/const.py
+++ b/pyvex/const.py
@@ -3,7 +3,7 @@ from . import VEXObject
 # IRConst hierarchy
 class IRConst(VEXObject):

-    __slots__ = ['tag', 'value']
+    __slots__ = ['value']

     type = None
     tag = None
diff --git a/pyvex/expr.py b/pyvex/expr.py
index 680be83..b8c58d9 100644
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -5,8 +5,6 @@ class IRExpr(VEXObject):
     IR expressions in VEX represent operations without side effects.
     """

-    __slots__ = ['tag']
-
     tag = None

     def __init__(self):
diff --git a/pyvex/stmt.py b/pyvex/stmt.py
index e2d95b6..3a69020 100644
--- a/pyvex/stmt.py
+++ b/pyvex/stmt.py
@@ -6,8 +6,6 @@ class IRStmt(VEXObject):
     IR statements in VEX represents operations with side-effects.
     """

-    __slots__ = ['tag']
-
     tag = None

     def __init__(self):
```

rhelmot:
((pushed something to the angr repo to make test_dfg pass now that expressions don't have an arch on them))

rhelmot:
```diff
diff --git a/pyvex/stmt.py b/pyvex/stmt.py
index 3a69020..2559d96 100644
--- a/pyvex/stmt.py
+++ b/pyvex/stmt.py
@@ -156,11 +156,11 @@ class PutI(IRStmt):

     @staticmethod
     def _from_c(c_stmt):
-        return PutI(IRRegArray(c_stmt.Ist.PutI.details.descr),
+        return PutI(IRRegArray._from_c(c_stmt.Ist.PutI.details.descr),
                     IRExpr._from_c(c_stmt.Ist.PutI.details.ix),
-                    IRExpr._from_c(c_stmt.Ist.PutU.details.data),
+                    IRExpr._from_c(c_stmt.Ist.PutI.details.data),
                     c_stmt.Ist.PutI.details.bias)
-
+
 class WrTmp(IRStmt):
     """
     Assign a value to a temporary.  Note that SSA rules require each tmp is only assigned to once.  IR sanity checking
```

mattrepl:
Ok, thanks for the cleanup. If you guys prefer that, it's fine by me. Just for reference, it looks like `ffi.gc()` may have worked as well.

http://cffi.readthedocs.io/en/latest/ref.html#ffi-gc

rhelmot:
Alright, after a few more changes, all tests pass. Will merge soon.

ltfish:
@mattrepl Thank you for this PR. Those changes are mostly great. I do, however, have different views about moving `arch` away from `IRStmt`s and `IRExpr`s. That change makes pretty-printing register names impossible when I only have a reference to an `IRStmt` or an `IRExpr`, without any reference to an `IRSB`. It happens a lot when I'm debugging the static analysis code, which drives me crazy. To this end, I'd like to revert to the old way that `arch` is passed down to `IRStmt`s and `IRExpr`s. However, it is optional, so you don't have to pass `arch` if you don't need register name pretty printing. What do you think?

mattrepl:
The general idea here was to remove passing down more information than is necessary. The `arch` attribute was only used for getting architecture-specific register names. If we really need to pass something  through as an attribute, it should probably just be a `reg_name` variable and not `arch` and `tyenv`. (Just my opinion, I'm a random dude on github ;)

If it's not too painful, a (perhaps cleaner?) way to do this is to write a little helper function/closure that knows the architecture and tyenv you're working with and will call `__str__` with the optional `reg_name` parameter set.

`
def make_printer(arch, tyenv):
  def pp_str(stmt):
    if isinstance(stmt, pyvex.stmt.WrTmp):
      return stmt.__str__(reg_name=arch.translate_register_name(stmt.offset, stmt.data.result_size(tyenv)/8)
    # And so on...
  return pp_str
`

The nastiest part about that is the `pp_str` closure requires the `tyenv` which is block-specific. But maybe that can be abstracted away for your use case?

ltfish:
> If it's not too painful, a (perhaps cleaner?) way to do this is to write a little helper function/closure that knows the architecture and tyenv you're working with and will call __str__ with the optional reg_name parameter set.

I'm already doing this in angr code, both during debugging and in statement printing. It makes me feel that the logic (making `statement` know `arch`, which is something it should have known already after creation) is unnecessarily "leaked" from PyVEX to angr, or any other project that uses it.

I definitely agree with you that not having `arch` there is way cleaner, but that change makes using `IRStmt`s and `IRExpr`s (more specifically, pretty-printing the register names) harder than before. While having clean code is a must, I don't think impacting how easy it is to use angr, as well as the underlying libraries, is the way to go.

Personally, I can always monkey-patch my local version of PyVEX and make it pretty-printing the register names when debugging static analyses. However, in this case, I believe passing `arch` around, which slightly reduces the cleanness of the code, will save some keystrokes and benefit users more.

Just my two cents :-)

> The nastiest part about that is the `pp_str` closure requires the `tyenv` which is block-specific

I might be wrong on this point, but for register name pretty-printing, `tyenv` (or the `result_size`) is not useful. The offset itself is enough to pinpoint a specific register.

mattrepl:
> It makes me feel that the logic (making statement know arch, which is something it should have known already after creation) is unnecessarily "leaked" to angr from PyVEX, or to any other project that uses it.

Yeah, I wavered on that myself. But I felt the cleanliness combined with following libVEX's example and still supporting pretty-printing of statements and expressions at the `IRSB` level were convincing reasons to make the change. How often is it that angr users or devs print lifted statements and expressions without printing the entire block? It's not often in my work, but maybe it is in yours and other's.

For general use of PyVEX, the user wouldn't expect to be required to include architecture information when libVEX doesn't require it. For all use cases other than printing out lifted code (granted, that's angr's main use case, for now), requiring extra info like `arch` is annoying.

> I might be wrong on this point, but for register name pretty-printing, tyenv (or the result_size) is not useful. The offset itself is enough to pinpoint a specific register.

Yes, having only the offset is enough to get a register name. But it would be the name used to refer the full-sized register. E.g., `rax` vs `eax`, `ax`, etc.

Issue #54, opened 2017-01-20
install fails on arch linux because "cl" command exists, but is something different


f0rki:
I'm having trouble getting the latest pyvex to work on arch linux. I'm using a virtualenv. Install works fine with pip. When I try to import pyvex, I get the following error:

```
$ python -c 'import pyvex'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/pyvex/__init__.py", line 31, in <module>
    pvc = _find_c_lib()
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/pyvex/__init__.py", line 23, in _find_c_lib
    ffi.cdef(_ffi_str)
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/cffi/api.py", line 105, in cdef
    self._cdef(csource, override=override, packed=packed)
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/cffi/api.py", line 119, in _cdef
    self._parser.parse(csource, override=override, **options)
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/cffi/cparser.py", line 299, in parse
    self._internal_parse(csource)
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/cffi/cparser.py", line 304, in _internal_parse
    ast, macros, csource = self._parse(csource)
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/cffi/cparser.py", line 262, in _parse
    self.convert_pycparser_error(e, csource)
  File "~/.virtualenvs/pyvex/lib/python2.7/site-packages/cffi/cparser.py", line 291, in convert_pycparser_error
    raise api.CDefError(msg)
cffi.api.CDefError: cannot parse "extern VexControl vex_control;"
:3:19: before: vex_control
```

Apparently the file `_ffi_str` variable contains only the string "extern VexControl vex_control;". 

After some digging I came to the conclusion that the issues is in [make_ffi.py#L58](https://github.com/angr/pyvex/blob/master/make_ffi.py#L58). The first tried command is `cl`. This is a binary that does something completely different (apparently part of the cliquer package) on my system.
This fails silently as there is no check for `Popen.returncode`. Maybe using `subprocess.check_output` would be better suited here, since it throws an exception on non-zero exit status?

[...]

PR #55, opened 2017-01-23
Check returncode and stderr when running compiler in make_ffi.py

f0rki:
This fixes the silent install fail when 'cl' is existing, but is not a compiler. 
Also uses the `CXX` environment variable if it's set.

fixes issue angr/pyvex#54

I have a similar patch for angr/simuvex#70 ready.

### DIFF
--- a/make_ffi.py
+++ b/make_ffi.py
@@ -54,17 +54,33 @@ def find_good_scan(known_good, questionable):
     return find_good_scan(known_good, questionable[:fail-2-len(known_good)])
 
 def doit(vex_path):
-    #vex_pp,_ = subprocess.Popen(['cpp', 'vex/pub/libvex.h'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT).communicate()
-    cmd1 = ['cl', '-I'+vex_path, 'pyvex_c\\pyvex.h', '-E']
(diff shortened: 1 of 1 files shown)


zardus:
This looks ok to me, and the testcases still work (except for the known-broken pyvex testcases), but a second look might be nice. @rhelmot, would you mind taking a look?

make_ffi.py
```diff
@@ -54,17 +54,33 @@ def find_good_scan(known_good, questionable):
     return find_good_scan(known_good, questionable[:fail-2-len(known_good)])
 
 def doit(vex_path):
-    #vex_pp,_ = subprocess.Popen(['cpp', 'vex/pub/libvex.h'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT).communicate()
-    cmd1 = ['cl', '-I'+vex_path, 'pyvex_c\\pyvex.h', '-E']
-    cmd2 = ['cpp', '-I'+vex_path, 'pyvex_c/pyvex.h']
-    for cmd in (cmd1, cmd2):
+    cxxlist = ['cl', 'cpp']
+    cxx = os.getenv("CXX")
```
rhelmot: traditionally, the env var used for a preprocessor is CPP

rhelmot: https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

PR #57, opened 2017-01-31
Fixed incorrect attribute name in IRSB construction method.

mattrepl:
Typo with a simple fix in the `IRSB.from_py` method.


zardus:
@subwire, does this interfere with your current stuff? If not, I'll merge it, but if yes, we should figure that out more carefully.

rhelmot:
I think this is a pure bugfix, shouldn't hurt anything.

zardus:
Works for me :-)

Issue #56, opened 2017-01-25
Build of libVEX fails on Mac OS X with default Clang


mattrepl:
The fix for #53 added the -malign-double compiler flag. Unfortunately this flag is not supported in the version of Clang (Apple LLVM version 8.0.0 (clang-800.0.42.1)) that ships with Mac OS X Sierra (10.12.3).

Current workarounds are either installing a newer version of Clang that supports the flag or to create a `vex` directory (copy `vex-master` in `pyvex/`) and removing the `-malign-double` flag from `Makefile-gcc`.

rhelmot:
FYI: this has been fixed on my current development branch, but that branch is also a gigantic rebase of vex on upstream and a cleanup of our git history in preparation of submitting our changes to valgrind, so it's not getting pushed for a few days...

vancaho:
@mattrepl Have you installed pyvex on Mac OSx Sierra (10.12.3) successfully? I have the same problem with you, but after removing the -malign-double flag, another error is reported:  File "/pyvex/make_ffi.py", line 92, in doit
    raise Exception("Couldn't process pyvex headers - set CPP env var")
Could you please give me some good suggestion? Thanks

rhelmot:
That error message you are seeing was actually added more recently than this issue was opened: https://github.com/angr/pyvex/pull/55/files

I would recommend enabling logging so you can see the error messages being logged.

shelvenzhou:
@vancaho same problem with you
just revert the change introduced by
https://github.com/angr/pyvex/pull/55/files
and everything will be okay

rhelmot:
@f0rki your change is causing issues on os x, do you have any insight to why that might be?

f0rki:
Strange. Maybe we should also add `-E` in case `CPP` points to another compiler binary (e.g. `clang` or `gcc`) to run the preprocessor?

@shelvenzhou can you import pyvex after reverting and installing? What is the `cpp` binary on your system? Is it Can you execute `python make_ffi.py ../vex/pub` (or wherever your vex master is) script manually? then you should get debug output and see what has failed.

shelvenzhou:
@f0rki yes everything works properly.

Here's the output of my `cpp -v`:

> Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin16.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
[...]

PR #58, opened 2017-02-10
Fix cpp call on mac

f0rki:
see discussion in #56


rhelmot:
@fish it looks like you pushed something that conflicts with this very very recently, in make_ffi.py. Could you review this?

f0rki:
Btw. checking if stderr contains some output, might not have been such a good idea, because I think a compiler might print warnings to stderr, right?

rhelmot:
It looks like whatever issues were preventing this from being merged before have been resolved. Thanks!

PR #70, opened 2017-04-10
Update pyvex.def, fix msvc noreturn directive

0xbc:
Builds were broken again under VS2015 - this patch should fix the build, but also depends on a companion PR I'll be sending through to `vex` shortly.

### DIFF
--- a/pyvex_c/pyvex.c
+++ b/pyvex_c/pyvex.c
@@ -53,7 +53,11 @@ jmp_buf jumpout;
 //
 //======================================================================
 
+#ifdef _MSC_VER
+__declspec(noreturn)
+#else
 __attribute__((noreturn))
+#endif
 static void failure_exit(void) {
 	longjmp(jumpout, 1);
 }
--- a/pyvex_c/pyvex.def
+++ b/pyvex_c/pyvex.def
@@ -1,12 +1,8 @@
 LIBRARY pyvex.dll
 
 EXPORTS
-  vex_block_bytes
-  vex_block_inst
   vex_control
   vex_init
-  enable_debug
-  last_error
   emptyIRTypeEnv
   emptyIRSB
   IRConst_U1
--- a/pyvex_c/pyvex.def
+++ b/pyvex_c/pyvex.def
@@ -3,6 +3,7 @@ LIBRARY pyvex.dll
 EXPORTS
   vex_control
   vex_init
+  vex_lift
   emptyIRTypeEnv
   emptyIRSB
   IRConst_U1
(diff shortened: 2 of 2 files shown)


rhelmot:
hey so; by any chance can I get you to debug our windows CI for us? I merged this and reran an old CI build that had failed on building libvex and it still failed on building libvex

https://ci.appveyor.com/project/rhelmot/claripy/build/1.0.82

Is this something that you can fix easily?

0xbc:
Happy to try and help!

From a quick glance, my suggestion would be that the problem is the use of `inline` and C99 support. appveyor looks like it's using VS2013 there, but I've been using VS2015 (which I figured would be the MSVC baseline for building `angr` - anything earlier and the C99 support just isn't good enough and breaks everything). Can you configure the CI to use VS2015 instead?

0xbc:
In https://github.com/angr/claripy/blob/master/appveyor.yml, it looks like you've specified the VS2015 image, but then you've hardcoded the path to VS2012 in the install script.

rhelmot:
I do want to use VS 2013, just for maximum compatibility and such. I seem to recall there being an issue with appveyor that forced me to use their VS 2015 image for whatever reason, but it also has VS 2013 installed :P

I know we've successfully built under VS 2013 in the past...

rhelmot:
You know what, fuck it, let's just use VS 2015

0xbc:
That's the spirit! :smile:

I was gonna say - I don't think `inline` would ever have compiled in VS2013 - cf. https://msdn.microsoft.com/en-us/library/z8y1yy88(v=vs.120).aspx.

> The inline keyword is available only in C++. The __inline and __forceinline keywords are available in both C and C++. For compatibility with previous versions, _inline is a synonym for __inline.

rhelmot:
Yeah what we did at one point was to either #define it out or to #define it to _inline, but if there are other issues with additional c99 stuff I'd rather not go down that route

0xbc:
I was previously trying to use VS 2012 and there were a buttload of C99 issues. I don't have an install of VS 2013 handy so I'm not sure how many other problems there are (although I could get my hands on one to test it if you really wanted to try and stick with it, would just take me a bit longer).

rhelmot:
to clarify - I don't think there is any such thing as VS 2012. there's a weird naming scheme issue where there's Visual Studio 2013, which bundles MSVC 12.0, and then Visual Studio 2015, which bundles MSVC 14.0. If you thought you were using VS 2012 you were probably using VS 2013

rhelmot:
I appear to be... wrong about this? I am confused

0xbc:
No... I think you're correct, sorry - I must have been using VS 2012, in which case I can confirm, yes there are a myriad of C99 issues building pyvex, definitely avoid it if you can.

0xbc:
Wait.. no, I definitely have VS 2012. So confused... anyway, the point is, let's use VS2015. 😸

PR #71, opened 2017-04-13
fixed print statements in setup.py

jacobj10:
for python3 compatibility


rhelmot:
...angr is a VERY long way off from python 3 compatibility, but thanks for the gesture!

rhelmot:
yeah, like, even just pyvex is blisteringly incompatible with python 3, for lots of reasons, the first of which that crops up is the use of the `urllib3` module. If you want pyvex to work in py3k, you have some work ahead of you...

jacobj10:
Thanks for the quick response!

PR #77, opened 2017-06-06
Fix ARM post-processing function to determine jumpkind

tyb0807:
Fixing ARM wrong jumpkind when 'lr' is affected as in angr/angr#307.
Thanks @ltfish for your help.


pyvex/lift/fixes.py
```diff
@@ -17,19 +17,42 @@ def _post_process_ARM(self):
...
+                        if isinstance(stt.data, expr.RdTmp):
+                            pc_holders.append(stt.data.tmp)
+                        # ... by a constant, so test whether it is the address of the
+                        # next instruction
+                        elif isinstance(stt.data, expr.Const):
+                            if stt.data.con == next_inst_addr:
+                                lr_store_pc = True
+                            break
+                # Tracking down the temporary variable that modifies lr
+                if isinstance(stt, stmt.WrTmp) and stt.tmp in pc_holders:
```
ltfish: You can make it an `elif` instead of a new `if` statement.

ltfish:
I believe you are missing at least one case: A temporary variable can be assigned to by another temporary variable before used in a `Put` IR statement. For example:

```
t29 = 0x400458
t30 = t29
Put(PC) = t30
```

I don't think your code can handle this case currently.

In addition, can you please provide some test cases for this new logic?

pyvex/lift/fixes.py
```diff
@@ -17,19 +17,42 @@ def _post_process_ARM(self):
...
+                    # pc found
+                    if isinstance(stt.data, expr.Get):
+                        if stt.data.offset == self.irsb.arch.registers['pc'][0]:
+                            lr_store_pc = True
+                            break
+                    # the temporary variable can hold not exactly pc but
+                    # something relative to pc
```
tyb0807: I think this part of code should handle the case you mention.

pyvex/lift/fixes.py
```diff
@@ -17,19 +17,42 @@ def _post_process_ARM(self):
...
+                    elif isinstance(stt.data, (expr.Unop, expr.Binop,
+                                               expr.Triop, expr.Qop)):
+                        pc_holders.pop()
+                        for arg in stt.data.args:
+                            if isinstance(arg, expr.RdTmp):
+                                pc_holders.append(arg.tmp)
```
tyb0807: I think this part of code should handle the case you mention

pyvex/lift/fixes.py
```diff
@@ -17,19 +17,42 @@ def _post_process_ARM(self):
...
+                if isinstance(stt, stmt.WrTmp) and stt.tmp in pc_holders:
+                    # pc found
+                    if isinstance(stt.data, expr.Get):
+                        if stt.data.offset == self.irsb.arch.registers['pc'][0]:
+                            lr_store_pc = True
+                            break
+                    # the temporary variable can hold not exactly pc but
+                    # something relative to pc
+                    elif isinstance(stt.data, (expr.Unop, expr.Binop,
+                                               expr.Triop, expr.Qop)):
```
tyb0807: I think this part of code should handle the case you mention

tyb0807:
@ltfish Here you are!

ltfish:
I'm OK with taking this PR. @zardus @rhelmot Any comments?

zardus:
I'm a bit worried about performance impact, but I'm ok with this PR. The fact that it has some testcases makes it extremely tempting, too :-)

tyb0807:
So what do you think I should change to improve performance? My first attempt was to traverse all statements in reverse order, so it can check whether the block modifies LR within the last 2 instructions, but since I have to forward execute the block to compute the value being written to LR, I traverse all statements from the beginning of the block. Maybe you want to do it in 2 passes? Reverse order first then in-order?

ltfish:
> So what do you think I should change to improve performance?

Let it be. We'll notice the degraded performance once it becomes a real issue in the future. We can then profile and solve it.

I'm going to merge this PR.

Issue #78, opened 2017-06-29
ffi.cdef parsing causes build failure on PyPy 5.8 virtualenv


Qwaz:
## Environment
PyPy 5.8 on virtualenv
Using the latest commit `a87324f`

## Description
PyPy 5.8 uses different error message format for cffi.CDefError to Python 2.7.13, which causes build failure.

Python 2.7.13 error format example
```
cannot parse "x &= 0xFFFFFFFFLL;"
:34:6: before: &=
```

PyPy 5.8 error format example
```
cannot parse "Int r = (x == 0) ? ((Bool)0) : ((Bool)1);"
<cdef source string>:31:30: before: 0
```
```
<cdef source string>:32: unexpected <FuncDef>: this construct is valid C but not valid in cdef()
```

The second error format causes out of bound exception, because it has two `:` but only one line.
https://github.com/angr/pyvex/blob/a87324f88aa621436d4b8077b828b92ce51e08c2/make_ffi.py#L41-L42

Qwaz:
Python 2.7.13 ffi error message gives different line numbers even if the prefix of the code is same.
```
------------------------------------
failed at 34, known_good: 31, questionable: 1386
 30| typedef unsigned char Bool;
 31|    Int r = (x == 0) ? ((Bool)0) : ((Bool)1);
 32|    return (Bool)r;
 33| }
 34| static inline UChar toUChar ( Int x ) {
 35|    x &= 0xFF;
 36|    return (UChar)x;
 37| }
cannot parse "return (Bool)r;"
:35:4: before: return
------------------------------------
------------------------------------
failed at 33, known_good: 31, questionable: 1386
 29| typedef double Double;
 30| typedef unsigned char Bool;
 31|    Int r = (x == 0) ? ((Bool)0) : ((Bool)1);
 32|    return (Bool)r;
 33| }
cannot parse "return (Bool)r;"
:34:4: before: return
------------------------------------
```

PR #79, opened 2017-07-03
Change ffi parsing algorithm

Qwaz:
This commit tries to fix issue #78

## Changes
* Remove unnecessary `find_good_bsearch` function
* Change the code format to fit PEP 8, except the line length rule
* Improve ffi parsing speed by changing the parsing algorithm

## TODO
New algorithm adds extra `Int r = (x == 0) ? ((Bool)0) : ((Bool)1);` line to `vex_ffi.py`. The line itself is a valid Python FFI definition, but it is in a function body and should be removed from the result.

### DIFF
--- a/make_ffi.py
+++ b/make_ffi.py
@@ -9,50 +9,76 @@ import logging
 l = logging.getLogger('cffier')
 l.setLevel(logging.DEBUG)
 
-def find_good_bsearch(known_good, questionable):
-    ffi = cffi.FFI()
-    l.debug("bsearch - trying %d good and %d questionable", len(known_good), len(questionable))
 
-    try:
-        ffi.cdef('\n'.join(known_good + questionable))
-    except (cffi.CDefError, AssertionError):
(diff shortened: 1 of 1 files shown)


Qwaz:
Verified building on Linux, PyPy 5.8 and Python 2.7.13 by `python setup.py build`

zardus:
@ltfish, is the CI failure (https://travis-ci.org/angr/pyvex/jobs/249566620) expected? Seems to be an issue with CLE usage.

rhelmot:
I fixed that error on the end times already!

zardus:
Yeah, I thought I saw something about that. In that case, I've pulled the trigger! Thanks, @Qwaz!

Qwaz:
As I mentioned in TODO section, this commit adds extra line to `vex_ffi.py`(which won't be harmful, but neither correct). I'll come up with another PR when I fix it. ;)

PR #92, opened 2017-09-28
Expanded Reality support, Gymrat gets more swole, better imagination support

subwire:
The second phase of our journey into other realms of computational existence.

Here's what's new:
- Better support for op and type imagination.  You can now / should now ask pyvex for the types of ops.
- Move gymrat to a two-pass model, enabling lifted instructions to be aware of the decoded instruction stream.  This allows certain instructions that, for example, must skip an instruction, to know how long the skipped instruction is without a CCall.  This also avoids the BF bracket-matching CCall problem.
- Make VexValue more flexible, adding the "r" ops, and getitem.

### DIFF
--- a/pyvex/lift/util/instr_helper.py
+++ b/pyvex/lift/util/instr_helper.py
@@ -1,6 +1,6 @@
 from lifter_helper import ParseError
 from syntax_wrapper import VexValue
-from vex_helper import JumpKind, IRExpr
+from vex_helper import JumpKind, IRExpr, vex_int_class
 
 import abc
 import string
@@ -105,9 +105,10 @@ class Instruction:
         # Then do the actual stuff.
         inputs = self.fetch_operands()
         retval = self.compute_result(*inputs)
+        vals = list(inputs)
         if retval is not None:
-            vals = list(inputs) + [retval]
-            self.commit_result(retval)
+            vals.append(retval)
+            self.commit_result(*vals)
         self.compute_flags(*vals)
 
     def commit_result(self, *args):
@@ -213,7 +214,8 @@ class Instruction:
         :param ty: The type of the resulting VexValue
         :return: a VexValue
         """
-        assert not (isinstance(val, VexValue) or isinstance(val, IRExpr))
+        if isinstance(val, VexValue) and not isinstance(val, IRExpr):
+            raise Exception('Constant cannot be made from VexValue or IRExpr')
         rdt = self.irsb_c.mkconst(val, ty)
         return VexValue(self.irsb_c, rdt)
 
@@ -262,7 +264,9 @@ class Instruction:
         aren't normal jumps (e.g., calls, interrupts, program exits, etc etc)
         :return: None
         """
-        assert isinstance(to_addr, VexValue)
+        if not isinstance(to_addr, VexValue):
+            to_addr_ty = vex_int_class(self.irsb_c.irsb.arch.bits).type
+            to_addr = self.constant(to_addr, to_addr_ty) # TODO archinfo may be changing
         if not condition:
             # This is the default exit.
             self.irsb_c.irsb.jumpkind = jumpkind
--- a/pyvex/lift/util/instr_helper.py
+++ b/pyvex/lift/util/instr_helper.py
@@ -78,7 +78,7 @@ class Instruction:
         self.lift(irsb_c, past_instructions, future_instructions)
 
     def mark_instruction_start(self):
-        self.irsb_c.imark(self.addr, self.bytewidth, self.bytewidth)
+        self.irsb_c.imark(self.addr, self.bytewidth, 0)
 
     def fetch_operands(self):
         """
--- a/pyvex/lift/util/instr_helper.py
+++ b/pyvex/lift/util/instr_helper.py
@@ -62,20 +62,19 @@ class Instruction:
     data = None
     irsb_c = None
 
-    def __init__(self, irsb_c, bitstrm, addr):
+    def __init__(self, bitstrm, endianness, addr):
         """
         Create an instance of the instruction
         :param irsb_c: The IRSBCustomizer to put VEX instructions into
         :param bitstrm: The bitstream to decode instructions from
         :param addr: The address of the instruction to be lifted, used only for jumps and branches
         """
-        self.irsb_c = irsb_c
         self.addr = addr
         self.width = len(self.bin_format)
-        self.data = self.parse(bitstrm)
+        self.data = self.parse(bitstrm, endianness)
 
-    def __call__(self):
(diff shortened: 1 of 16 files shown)


pyvex/lift/util/instr_helper.py
```diff
@@ -168,10 +180,14 @@ def parse(self, bitstrm):
...
             data = self._extra_parsing(data, bitstrm)
-        afterbits = bitstrm.bitpos
-        self.bitwidth = afterbits - beforebits
         return data
 
+    @property
+    def bytewidth(self):
+        if self.bitwidth % self.arch.byte_width != 0:
+            raise ValueError("Instruction is not a multiple of bytes wide!")
+        return self.bitwidth / self.arch.byte_width
```
rhelmot: Use the floor division // operator here

pyvex/lift/util/syntax_wrapper.py
```diff
@@ -124,31 +126,59 @@ def __add__(self, right):
...
     @checkparams
     @vvifyresults
     def __truediv__(self, right): # Note: nonprimitive
         return self / right
-
+    """
+    @checkparams
+    def __rtruediv__(self, left):
+        return left.__truediv__(self)
+    """
```
rhelmot: generally we don't like using strings for multiline comments. just stick a # in front of them. If you're using a good editor this should be easy.

pyvex/lift/util/syntax_wrapper.py
```diff
@@ -206,14 +236,25 @@ def __mod__(self, right): # Note: nonprimitive
         return self.rdt - (self.rdt / right.rdt) * right.rdt
 
     @checkparams
+    def __rmod__(self, left):
+        return left % self
```
rhelmot: are you sure this works? this just seems like it'll infinitely recurse

pyvex/lift/util/syntax_wrapper.py
```diff
@@ -206,14 +236,25 @@ def __mod__(self, right): # Note: nonprimitive
...
     def __mul__(self, right):
-        return self.irsb_c.op_mull(self.rdt, right.rdt, signed=self.is_signed)
+        if self._is_signed:
+            return self.irsb_c.op_smul(self.rdt, right.rdt)
+        else:
+            return self.irsb_c.op_umul(self.rdt, right.rdt)
+
+    @checkparams
+    def __rmul__(self, left):
+        return left * self
```
rhelmot: see above re: infinite recursion

pyvex/lift/util/syntax_wrapper.py
```diff
@@ -124,31 +126,59 @@ def __add__(self, right):
...
-    def __sub__(self, right): # I'm not sure if we want rsub
+    def __sub__(self, right):
         return self.irsb_c.op_sub(self.rdt, right.rdt)
-
+    """
+    @checkparams
+    def __rsub__(self, left):
+        # TODO: FIXME
+        return left - self
+    """
```
rhelmot: see below re: multiline comments

pyvex/lift/util/syntax_wrapper.py
```diff
@@ -124,31 +126,59 @@ def __add__(self, right):
...
-
+        if self._is_signed:
+            return self.irsb_c.op_sdiv(self.rdt, right.rdt)
+        else:
+            return self.irsb_c.op_udiv(self.rdt, right.rdt)
+    """
+    @checkparams
+    def __rdiv__(self, left):
+        return left / self
+    """
```
rhelmot: see below re: multiline comments

pyvex/lift/util/syntax_wrapper.py
```diff
@@ -124,31 +126,59 @@ def __add__(self, right):
...
     @checkparams
     @vvifyresults
     def __floordiv__(self, right): # Note: nonprimitive
         return self / right
-
+    """
+    @checkparams
+    def __rfloordiv__(self, left):
+        return left // self
+    """
```
rhelmot: see below re: multiline comments

pyvex/lift/util/syntax_wrapper.py
```diff
@@ -124,31 +126,59 @@ def __add__(self, right):
...
+        return left.__truediv__(self)
+    """
     @checkparams
     @vvifyresults
     def __and__(self, right):
         return self.irsb_c.op_and(self.rdt, right.rdt)
 
     @checkparams
+    def __rand__(self, left):
+        return left & self
```
rhelmot: see below re: infinite recursion

rhelmot:
So the infinite recursion stuff I was talking about is in fact mitigated by the `checkparams` wrapper - the arguments are converted into VexValues and when they hit the reverse op function, which just performs the exact same operation over again, this time both are VexValues and the forward op happens. The infinite recursion behavior can be seen when you try to use a datatype which can't be handled by the converter:

```python
In [14]: vv = VexValue.Constant(bbc, 42, 'Ity_I32')

In [25]: () - vv
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-25-3fea6a352d10> in <module>()
----> 1 () - vv

/home/audrey/angr/pyvex/pyvex/lift/util/syntax_wrapper.py in inner_decorator(self, *args, **kwargs)
     19                     args[args.index(arg)] = VexValue.Constant(self.irsb_c, arg, thetype)
     20             args = tuple(args)
---> 21             return fn(self, *args, **kwargs)
     22         return inner_decorator
     23     return decorator

/home/audrey/angr/pyvex/pyvex/lift/util/syntax_wrapper.py in __rsub__(self, left)
    118     @checkparams()
    119     def __rsub__(self, left):
--> 120         return left - self
    121
    122     @checkparams()

... last 2 frames repeated, from the frame below ...

/home/audrey/angr/pyvex/pyvex/lift/util/syntax_wrapper.py in inner_decorator(self, *args, **kwargs)
     19                     args[args.index(arg)] = VexValue.Constant(self.irsb_c, arg, thetype)
     20             args = tuple(args)
---> 21             return fn(self, *args, **kwargs)
     22         return inner_decorator
     23     return decorator

RuntimeError: maximum recursion depth exceeded
```

So like.... this works, but it feels extremely EXTREMELY skeezy to me. If you really don't think it's a problem I'll merge it if you add some type assertions to prevent the above from happening by accident, but `def __rsub__(self, other): return other - self` is a HUGE red flag for anyone who knows python.

sorry for being obnoxious about this it's just very important code

rhelmot:
Oh and one last thing - you should make sure that all classes inherit from `object` so they become what are called "new style classes" and behave like this is python2.7 instead of python2.2

PR #98, opened 2017-11-06
Exit statement should get the offset of IP in register files instead of the block address.

ltfish:


subwire:
Let's double-check on this one first.  We were using the lift-time instruction address as IP to avoid complications with SimEngineVEX's inaccurate IP.  I bet there's an inaccuracy because of this, but I'm curious, what is it?

ltfish:
I don't think we are on the same page. Supposedly the `offsIP` should be the offset of the IP register in VEX, e.g., for X86, the `offsIP` should always be `archinfo.arch_from_id('x86').registers['ip']`, which is 68. It is not the current instruction/block address.

subwire:
Revisited this; fixed it real quick to do exactly what you wanted, but take the work off the lifter author most of the time.

PR #100, opened 2017-11-06
Add a Mod operation to VEX.

ltfish:
It can be handy when implementing support for some architectures.

Requires [angr PR 745](https://github.com/angr/angr/pull/745).

### DIFF
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -721,7 +721,7 @@ def unop_signature(op):
     return size_type, (size_type,)
 
 def binop_signature(op):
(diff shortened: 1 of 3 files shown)


zardus:
The built-in VEX way of doing mod is Iop_DivMod(S|U) and extracting the remainder. Of course, this is easier (and leads to more analyzable IR), so I can see the argument for it (beyond just convenience).

subwire:
We have officially decided to open the pandora's box of adding new vex ops, so... let's do it!

PR #111, opened 2018-02-05
Small gymrat improvements

bennofs:
I decided to catch every `bitstring.ReadError`, alternatively we could also change each use site to convert this into a `ParserError`. But I think catching `ReadError` should be safe, according to the docs that is only thrown if you try to read past the end of the bitstring.

I also removed the `catch Exception, e: raise e` block, correct me if I'm wrong but I think that was redundant?

### DIFF
--- a/pyvex/lift/util/instr_helper.py
+++ b/pyvex/lift/util/instr_helper.py
@@ -89,7 +89,7 @@ class Instruction(object):
 
     def lift(self, irsb_c, past_instructions, future_instructions):
         """
-        THis is the main body of the "lifting" for the instruction.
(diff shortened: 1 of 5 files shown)


Nilocunger:
@bennofs The changes look great, thank you for contributing! One thing: I don't think that the modification to `set_bit` is necessary, since the `@check_params()` decorator should already convert any incoming `int`s into `VexValue`s.

bennofs:
@Nilocunger ok, the `check_params` decorator was broken. I added a new commit that removes the workaround and fixes `check_params`.

PR #113, opened 2018-02-09
Fix VexValue returning RdTmp on __mod__

Nilocunger:


bennofs:
What does the `note: nonprimitive` mean?

Nilocunger:
@bennofs It's something from a very early iteration of gymrat that I haven't deleted yet. I probably should

PR #114, opened 2018-02-09
instruction_helper: provide access to surrounding instructions

bennofs:
I introduced a new `apply_context` to avoid the possibility of a partially-modified irsb, for example the following would be bad, as we would be left with a half-finished IRSB and we cannot revert the partial changes:

```
def compute_result(...):
    self.put(...) # this modifies the IRSB
    # do something that finds out that we need more context
```

Introducing a new method for inspecting the context makes it harder to screw this up.

I included the index in the exception to allow further enhancments, for example we could find a way to deal with requiring access to more past context by lifiting earlier instructions. The index can then be used to figure out what we should do to provide the required context.

### DIFF
--- a/pyvex/lifting/util/instr_helper.py
+++ b/pyvex/lifting/util/instr_helper.py
@@ -1,10 +1,9 @@
-
 import abc
 import string
 import bitstring
 import logging
 
-from .lifter_helper import ParseError
+from .lifter_helper import ParseError, RequireContextError
 from .syntax_wrapper import VexValue
 from ...expr import IRExpr
 from .vex_helper import JumpKind, vex_int_class
@@ -12,6 +11,34 @@ from .vex_helper import JumpKind, vex_int_class
 
 l = logging.getLogger("instr")
 
+class ContextInstructions(object):
+    """Sequence of lookahead/lookback instruction context.
+
+    This sequence is list-like, but out-of-bounds raise a `RequireContextError`.
+    """
+    def __init__(self, available, past):
+        self._available = available
+        self._past = past
+
+    def __getitem__(self, key):
+        if isinstance(key, slice):
+            if key.start < 0:
+                raise ValueError("instructions context does not support negative index (slice start)")
+            if key.stop < 0:
+                raise ValueError("instructions context does not support negative index (slice stop)")
+
+            return (self._get(idx) for idx in xrange(key.start, key.stop, key.step or 1))
+        return self._get(key)
+
+    def _get(self, idx):
+        if idx < 0:
+            raise ValueError("instructions context does not support negative index")
+
+        if idx >= len(self._available):
+            bound = -idx if self._past else idx
+            raise RequireContextError(bound)
+
+        return self._available[idx]
 
 class Instruction(object):
     """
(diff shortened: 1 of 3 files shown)


Nilocunger:
@bennofs I like the idea, but I'm not sure this is the best way to do it. Instead of having an `apply_context` part of the lifting process, which could be unnecessarily restrictive, what would you think about doing something similar to the block lifting ([here](https://github.com/angr/pyvex/blob/master/pyvex/lift/__init__.py)) where we apply the instructions to empty IRSBs and then concatenate those IRSBs only if they don't need more context? That way the lifter writer doesn't need to be worried about accidentally corrupting the IRSB.

bennofs:
Oh, that sounds like a good idea! I'll try that out.

bennofs:
@Nilocunger implemented that. I had to rewrite the lifting method a little bit as it was getting too complex for me to understand. I think the new code should do the same as the old one, but I may have missed something.

bennofs:
For some reason, travis does not run pyvex tests. This is actually failing them, need to look into it.

bennofs:
Oh, they fail on master as well, that's why :/

bennofs:
Fixed on small issue with the jumpkind being set incorrectly and rebased. There are still some other issues regarding jumpkind and handling of max_bytes, but those exist on master as well so I will make a separate PR with fixes.

bennofs:
Rebased to adapt to the changes made on master.

pyvex/lifting/util/instr_helper.py
```diff
@@ -81,7 +108,9 @@ def __init__(self, bitstrm, arch, addr):
         self.data = self.parse(bitstrm)
 
     def __call__(self, irsb_c, past_instructions, future_instructions):
-        self.lift(irsb_c, past_instructions, future_instructions)
+        past  = ContextInstructions(past_instructions, True)
```
ltfish: Two spaces after `past`? Why?

pyvex/lifting/util/lifter_helper.py
```diff
@@ -86,21 +103,47 @@ def lift(self, disassemble=False, dump_irsb=False):
...
+
+            # if the block has a jumpkind, this is a full block so we can exit 
+            if self.irsb.jumpkind is not None:
                 break
-        else:
-            if len(irsb_c.irsb.statements) == 0:
-                raise LiftingException('Could not decode any instructions')
-            irsb_c.irsb.jumpkind = JumpKind.NoDecode
+
+        if len(self.irsb.statements) == 0:
```
ltfish: Would `if not self.irsb.statements` work? Pylint should complain about `len(...) == 0`.

pyvex/lifting/util/lifter_helper.py
```diff
@@ -86,21 +103,47 @@ def lift(self, disassemble=False, dump_irsb=False):
...
+            raise LiftingException('Could not lift any instructions')
+
+        if self.irsb.jumpkind is None:
+            if len(instructions) >= self.max_inst and self.max_inst is not None:
```
ltfish: I would swap the two conditions.

pyvex/lifting/util/lifter_helper.py
```diff
@@ -86,21 +103,47 @@ def lift(self, disassemble=False, dump_irsb=False):
...
         if disassemble:
             return [instr.disassemble() for instr in instructions]
-        self.irsb.jumpkind = JumpKind.Invalid
-        irsb_c = IRSBCustomizer(self.irsb)
+
         l.debug("Decoding complete.")
+        self.irsb.jumpkind = None
         for i, instr in enumerate(instructions[:self.max_inst]):
+            # first, create a scratch IRSB so we can throw the changes away if lifting fails
+            next_irsb_part = pyvex.IRSB.empty_block(self.irsb.arch, self.irsb.addr)
```
ltfish: `pyvex` is no longer imported. In fact, using `pyvex` inside the PyVEX package is not a good choice...

ltfish:
I read it through and left some comments. @Nilocunger any feedback?

bennofs:
@ltfish thanks for reviewing, fixed the issues.

ltfish:
Tests of `angr-platforms` failed. Can you please look into them?

bennofs:
@ltfish I think the test failure is caused by the issue (hopefully) fixed with this PR: https://github.com/angr/pyvex/pull/126

ltfish:
Huh, OK, I don't want to click the merge button without final words from @Nilocunger . I guess he is busy with his finals right now...

PR #117, opened 2018-02-15
Fix same RDT being used multiple times (now really)

bennofs:
CC @Nilocunger

The issue is that vex_helper sometimes gets passed raw RDTs (such as for set_bit) and then uses
the twice, so the copy in syntax_wrapper is not enough.

Not sure if the previous fix is now redundant with this change (probably)?
In that case, I can also revert the previous fix.


Nilocunger:
@bennofs Yeah, the previous fix is redundant. Please revert it and then we can merge this in.

PR #118, opened 2018-02-17
Fix tests

bennofs:
The tests should all pass now. I had to increase the "leeway" for the memory test to 5mb. Also, most of the tests are quite trivial now.


ltfish:
Thank you!

PR #121, opened 2018-02-25
Performance improvements in pyvex lifting

Nilocunger:


pyvex/block.py
```diff
@@ -114,52 +116,31 @@ def convert_expr(expr):
...
-                # This is the most frequent case
-                _handle_WrTmp(stmt)
-            else:
-                handler = _mapping.get(type(stmt), None)
-                if handler:
-                    handler(stmt)
-
-            # process all statements
+            stmttype = type(stmt)
+            if stmttype is WrTmp:
```
ltfish: Any good reason for us to switch back to cascading `if`s, instead of keep using the `dict`-based solution?

pyvex/lift/libvex.py
```diff
@@ -45,63 +45,24 @@ def lift(self):
...
-                    shortened_c_irsb = create_irsb(insts_shortage, bytes_shortage)
-
-                    self.irsb._from_c(shortened_c_irsb)
-                    c_irsb = shortened_c_irsb
-                    if self.irsb.size != extended_irsb.size:
-                        self.irsb.jumpkind = 'Ijk_NoDecode'
-                        self.irsb.next = 0
+            self.irsb.arch.vex_archinfo['hwcache_info']['caches'] = ffi.NULL
+            c_irsb = pvc.vex_lift(vex_arch,
+                                    self.irsb.arch.vex_archinfo,
```
ltfish: Indentation mismatch.

pyvex/block.py
Nilocunger: The cascading `if`s are shorter, easier to read, and faster.

PR #122, opened 2018-02-26
Khorben/netbsd

khorben:
This lets me build pyvex on NetBSD. It should not affect Windows, Linux or MacOS X.


rhelmot:
Looks good to me, thanks!

PR #123, opened 2018-02-26
Replace some regex matching with dict lookups to speed up lifting.

ltfish:
A 45% speed-up is observed on amd64 code.


ltfish:
... on CPython.

PR #126, opened 2018-03-08
More RdTmp copy fixes

bennofs:
There's other places where RdTmp's get used without going through `op_generic`. I hope I have now catched all of them. 

/cc @Nilocunger


zardus:
@Nilocunger, @subwire, could you take a look?

subwire:
Sorry for the delay! Looks ok to me.
Merging

PR #128, opened 2018-03-25
Py3k compatible

tyb0807:
Tests pass on my local branch. Let's see what Travis says


rhelmot:
All these changes look good. Thanks!

PR #130, opened 2018-05-01
Added Instruction.put_conditional

bob123456678:
Added a function to extend the gymrat lifter with high-level support for a conditional register put.

### DIFF
--- a/pyvex/lifting/util/instr_helper.py
+++ b/pyvex/lifting/util/instr_helper.py
@@ -261,6 +261,24 @@ class Instruction(object):
         """
         offset = self.lookup_register(self.irsb_c.irsb.arch, reg)
         self.irsb_c.put(val.rdt, offset)
+        
+        
+	def put_conditional(self, cond, valiftrue, valiffalse, reg):
+        """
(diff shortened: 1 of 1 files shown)


Nilocunger:
@bob123456678 Hey, welcome to contributing to angr! Before I look into merging this, could you explain why we need dedicated support in gymrat for conditional register put? It seems like users could write their own pretty short function if they have this need. Am I missing something?

bob123456678:
I recently wrote a (fairly complete) lifter for SH4 and found gymrat's high-level interface to be sufficient for (painlessly) implementing just about any instruction, however the lack of a conditional put was a notable exception.  It look a fair amount of digging through the source to figure out how to do this.  I feel like adding this to the core (a) would save other people time when implementing a lifter, as no knowledge of the underlying pyvex implementation would be necessary, and (b) is generic enough to be useful for many different architectures.

rhelmot:
Would someone actually responsible for this code like to merge this?

ltfish:
I'll hit merge as long as tests pass.

PR #131, opened 2018-05-04
The ARM Spotter (and the "gym")

subwire:
....because we don't skip arm day in this gym. Leg day is another matter.

This is the tracking PR for the ongoing effort to fill in the holes in VEX's support for weird things you don't find in userland ARM programs. Why? Firmware.

We also create a "gym" to hold the Gymrat-style lifters we plan to ship with Pyvex. This will probably not be the last.

Of course, this takes the form of a "Spotter" which will handle things libvex can't decode.
Unfortunately, all of those represent Super Messed Up Hardware Stuff(tm) which will only ever be really supported in some far-off fantasy dimension where all bytes are 9 bits, so we lift everything so far as a correct-sized no-op. We are building this iteratively as we load more weird stuff into angr and find NoDecodes, so if you're aware of more, please contribute.

Now supports Thumb, because we have to work on those video game muscles, yo

Checklist:
[ ] Testcases! Because TDD is like leg day, we probably shouldn't skip it
[ ] We might as well nail all the weird new Cortex instructions while we're at it.  There's an entire new FPU I doubt VEX can support today, let's try that one.
[ ] Rewrite the base Spotter testcases themselves to use this instead of the hack that's there now
[ ] Look into making all that Thumb-A stuff a Spotter too!
[ ] Add the real format for CPS, when we find an actual ARM document describing what it is.


pyvex/block.py
```diff
@@ -477,6 +477,23 @@ def _is_defaultexit_direct_jump(self):
         target = self.default_exit_target
         return target is not None
 
+    def py_collect_exits(self):
```
rhelmot: This should be private

pyvex/lifting/__init__.py
```diff
@@ -45,6 +45,12 @@ def lift(data, addr, arch, max_bytes=None, max_inst=None, bytes_offset=0, opt_le
 
     .. note:: If no instruction and byte limit is used, pyvex will continue lifting the block until the block
               ends properly or until it runs out of data to lift.
+
+    .. note:: If you specify an opt_level of 0, this will also disable many arch-specific optimizations
```
rhelmot: Could this actually be changed to opt_level=-1 to trigger the big unoptimization stuff?

subwire:
@ltfish adjusted base branch! 
Now it's ready to review.

@rhelmot regarding our py_.... hack, Fish will look and see if he can make that not necessary, if we keep it, i'll clean it up a bit

pyvex/block.py
ltfish: I will try to move this method to C instead.

pyvex/lifting/__init__.py
ltfish: I want to hear what you think about opt_level=-1...

PR #132, opened 2018-06-09
Fix CB{N}Z thumb instruction second attempt

ArtemShypotilov:
Added `strict_block_end` parameter everywhere in python vex lifting functionality, so it could be passed to `libvex.so`.

**WARNING**: This PR will break `angr/native/sim_unicorn.cpp` as it has call to `vex_lift`


rhelmot:
I had to turn the option off by default since it was causing some testcase failures, but it is there and it does work!

PR #135, opened 2018-06-24
Cache common IRExprs to reduce memory usage.

ltfish:


ltfish:
@subwire @Nilocunger What do you guys think of these changes? Now all the cached `IRExpr`s (`Const`, `RdTmp`, `U8`, `U16`, `U32`, and `U64`) are immutable.

pyvex/block.py
```diff
@@ -144,10 +145,10 @@ def convert_expr(expr):
             elif stmttype is CAS:
                 if stmt.oldLo not in invalid_vals: stmt.oldLo = convert_tmp(stmt.oldLo)
                 if stmt.oldHi not in invalid_vals: stmt.oldHi = convert_tmp(stmt.oldHi)
-            for e in stmt.expressions:
-                convert_expr(e)
+            for i in range(len(stmt.expressions)):
```
rhelmot: for i, _ in enumerate(stmt.expressions)?

rhelmot: for i, expr in enumerate(stmt.expressions): stmt.expressions[i] = convert_expr(expr)

pyvex/const.py
```diff
@@ -75,14 +80,19 @@ class U8(IRConst):
...
-        self.value = value
+        self._value = value
 
     def __str__(self):
         return "0x%02x" % self.value
 
     @staticmethod
     def _from_c(c_const):
-        return U8(c_const.Ico.U8)
+        global _U8_POOL
```
rhelmot: You don't need to mark this name as global if you don't modify the object it references

pyvex/const.py
```diff
@@ -94,14 +104,22 @@ class U16(IRConst):
...
-        return U16(c_const.Ico.U16)
+        global _U16_POOL
+
+        val = c_const.Ico.U16
+        if val < 1024:
+            return _U16_POOL[val]
+        return U16(val)
+
+
+_U16_POOL = [ U16(i) for i in range(1024) ]
```
rhelmot: maybe make it the constants -1024 through 1024, for here and below?

ltfish: They are unsigned... why do we want to have negative numbers?

rhelmot: Because vex will often encode subtracts as add with a large constant

Nilocunger:
Looks good to me!

PR #136, opened 2018-06-26
Remove tag setting, it causes an AttributeError

Kingloko:
Since the addition of slots to in commit 61efbdcf6303a936aa3de35011d2d1e3fe5fdea5 we can't create new instance properties anymore.


rhelmot:
Thanks!

PR #140, opened 2018-07-01
Implement expression replacing for IRStmt and IRExpr.

ltfish:
This commit fixes issue angr/angr#1096. Thanks @Alexeyan for reporting.

### DIFF
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -146,8 +146,15 @@ class IRSB(VEXObject):
             elif stmttype is CAS:
                 if stmt.oldLo not in invalid_vals: stmt.oldLo = convert_tmp(stmt.oldLo)
                 if stmt.oldHi not in invalid_vals: stmt.oldHi = convert_tmp(stmt.oldHi)
-            for i, expr in enumerate(stmt.expressions):
-                stmt.expressions[i] = convert_expr(expr)
+            # Convert all expressions
+            to_replace = { }
+            for expr in stmt.expressions:
+                replacement = convert_expr(expr)
+                if replacement is not expr:
+                    to_replace[expr] = replacement
+            for expr, replacement in to_replace.items():
+                stmt.replace_expression(expr, replacement)
+            # Add the converted statement to self.statements
             self.statements.append(stmt)
         extendwith.next = convert_expr(extendwith.next)
         self.next = extendwith.next
@@ -251,13 +258,12 @@ class IRSB(VEXObject):
     @property
     def expressions(self):
         """
-        A list of all expressions contained in the IRSB.
+        Return an iterator of all expressions contained in the IRSB.
         """
-        expressions = []
         for s in self.statements:
-            expressions.extend(s.expressions)
-        expressions.append(self.next)
-        return expressions
+            for expr in s.expressions:
+                yield expr
+        yield self.next
 
     @property
     def instructions(self):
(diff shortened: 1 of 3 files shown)


pyvex/block.py
```diff
@@ -251,13 +258,12 @@ def direct_next(self):
     @property
     def expressions(self):
         """
-        A list of all expressions contained in the IRSB.
+        Return an iterator of all expressions contained in the IRSB.
```
danse-macabre: I would strongly advise against making *any* property return an iterator, especially a public one. This would render property completely useless once it was accessed, as there's no way to tell, at what expression the generator has stopped. Not to mention, that once exhausted, the generator can't be run again.

ltfish: I guess I must be missing something here... Give me a little bit time for me to improve my Python knowledge... I thought returning generators is preferred to returning a list.

Alexeyan: I usually prefer lists.. Since they are easier to handle and more flexible. 
I don't always want to iterate over it.

Generators only make sense when iterating is the _only_ thing one could ever want to do with the entity.

ltfish: I'm trying to follow the rationale behind Python 3 changing `dict.items()` from a list to an iterator. By changing `.expressions` to an iterator, we don't always have to iterate through all expressions and create an entire list before returning. It saves time, especially in cases where not every item in `.expressions` will be used.

ltfish: But I think @danse-macabre is saying something else (this piece of code is wrong since it will only work once). I don't quite get it. Let me ask @rhelmot .

ltfish: @danse-macabre 

Sample code:
```
class A(object):
    def __init__(self, lst):
        self.lst = lst

    @property
    def expressions(self):
        for l in self.lst:
            yield l

a = A([1, 2, 3, 4, 5])
print list(a.expressions)
print list(a.expressions)
```

It prints out 
```
[1, 2, 3, 4, 5]
[1, 2, 3, 4, 5]
```
which is expected. The property `.expressions` can be used twice. Can you please elaborate what you mean in your comment?

danse-macabre: @ltfish Oh, my... I never thought it could work this way! I thought that the property would be the *generator* itself, rather than it would return a *generator*, hence it seemed to me, that the property would yield nothing, once exhausted.
It seems like my Python knowledge is far not as good as I thought :-) Sorry about the confusion.

Nevertheless, the code looks perfectly OK, except the two moments:
1) The length of the expressions list can't be obtained using the `len` function. 
2) IMHO, one might expect that the method/property that is supposed to return an iterator over items should be named like `iter_items`, or something. But this is, of course, deeply subjective.

ltfish: > The length of the expressions list can't be obtained using the `len` function.

Luckily, for `.expressions`, it's unlikely that you'll need the length, ever. If you need its length, you can still do `len(list(xxx.expressions))`. Doing this in the old way would cause the creation of quite a few lists (because both `.expressions` and `.child_expressions` were creating lists), and now the creation of those lists is gone for good.

> IMHO, one might expect that the method/property that is supposed to return an iterator over items should be named like `iter_items`, or something. But this is, of course, deeply subjective.

I totally agree with you on this point. However, Python 3 and some other packages (NetworkX 2, sortedcontainers 2) disagree with us. They are all removing traditional interfaces (like `.edges`) that used to return lists/sets/dicts, and are returning generators or iterable views when these methods or properties are called.

As part of our effort of migrating angr to Python 3, I'd like to follow the established norm.

PR #144, opened 2018-07-04
Consistent soft tabs in example README code

cub01d:


ltfish:
Can we use four spaces for each tab?

PR #156, opened 2018-09-14
Add S390X support

mephi42:


rhelmot:
Wow, is that really all it takes? Good stuff.

Do you have a sense of how many more pull requests you'll have before we have basic symbolic execution capabilities for s390x? we're gearing up for a major version bump in the next week or 2 (for purposes of switching to python 3), and this would be a great feature to have for that.

mephi42:
The major problem is that right now I can only analyze s390x on s390x: the VEX decoder supports only big-endian machines. I'm looking into changing that, but not sure how much time it would take.

rhelmot:
I've tweaked the lifter to deal with host/guest endness in a few places, here's a few examples:

https://github.com/angr/vex/commit/7084a7942a26ff1cf7de382bd77735e65524efb3
https://github.com/angr/vex/commit/3be7ed2c94f96eeea4f7cf04c2bf9525ef3cceb8
https://github.com/angr/vex/commit/62c35e6a3b013cc85b059c41bdb1d58c72f129c7

PR #160, opened 2018-10-02
Bugfix: prevent from raising exception when data type is str in python2

sashs:
If pyvex shall also support python2 this is necessary since this exception is always thrown when using python2.

```
    if isinstance(data, str):  <-- This is always true if str or bytes are used. 
        raise TypeError("Cannot pass unicode string as data to lifter")
    if isinstance(data, bytes):
```


sashs:
Oh sorry. :) I haven't looked at the website.

PR #171, opened 2018-10-22
Add test/test_s390x_exrl.py

mephi42:
Testcase for https://github.com/angr/vex/pull/20


rhelmot:
This is all fanstastic :) Can you add some smoketests for symbolic execution and static analysis in the angr repo? probably in test_accuracy and test_cfg*?

PR #175, opened 2018-11-20
a couple of bug fixes

canpadawan:
Hi,

a few bug fixes (well I hope :-)

- cbd4aa1: when using multiple lifters, first lifter may not decode anything and max_bytes may be None or decode everything and data_left would be b'' or ''. both cases were leading to exceptions
- 17b0181: `guest_stack_redzone_size` is asserted [here](https://github.com/angr/vex/blob/master/priv/guest_x86_toIR.c#L6786) but only set to 0 during lib initialization. If a block is lifted for AMD64 or PPC64 the value of `guest_stack_redzone_size` would be changed in `vex_prepare_vbi`, and then subsequent x86 block liftings hitting this assert would fail.

### DIFF
--- a/pyvex/lifting/__init__.py
+++ b/pyvex/lifting/__init__.py
@@ -161,7 +161,7 @@ def lift(data, addr, arch, max_bytes=None, max_inst=None, bytes_offset=0, opt_le
             data_left = data + final_irsb.size
         if max_inst is not None:
             max_inst -= final_irsb.instructions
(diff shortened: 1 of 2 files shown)


ltfish:
Just realized that we *never* pass `max_bytes=None` to the lifter. `max_bytes` will always set to 400 or another value that is less than 400 before passed in.

ltfish:
Thanks! I think they are legit fixes. Merging them in.

PR #177, opened 2018-12-20
Fix bugs in arm_post_processor_determine_calls()

danse-macabre:
The first commit fixes the bug in `tmps` and `regs` initialization code:

```c
	Addr tmps[MAX_TMP + 1] = { DUMMY };
	Addr regs[MAX_REG_OFFSET + 1] = { DUMMY };
```
In that code only the first element will get its value initialized to DUMMY, while all other elements will be initialized to zero. See https://stackoverflow.com/a/201116 for reference.

Interesting, that with the code above the following block gets its jumpkind set to Ijk_Call:

```
.text:0001A648                 LDR     R3, [SP,#0x180+var_68]
.text:0001A64C                 LDR     LR, =(sub_17C3C - 0x1A65C)
.text:0001A650                 CMP     R3, #2
.text:0001A654                 ADD     LR, PC, LR      ; sub_17C3C
.text:0001A658                 BEQ     loc_1AF50
```
... since the LDR instruction is ignored by the postprocessor and the addition at 0x1a654 results in
`LR = PC + 0`.

The second commit ensures that at the end of the computation the lr_store_pc variable will be set if and only if it contains the next_irsb_addr.

### DIFF
--- a/pyvex_c/postprocess.c
+++ b/pyvex_c/postprocess.c
@@ -90,11 +90,15 @@ void arm_post_processor_determine_calls(
 					IRConst *con = stmt->Ist.Put.data->Iex.Const.con;
 					if (get_value_from_const_expr(con) == next_irsb_addr) {
 						lr_store_pc = 1;
+					} else {
+						lr_store_pc = 0;
 					}
 				} else if (stmt->Ist.Put.data->tag == Iex_RdTmp) {
 					Int tmp = stmt->Ist.Put.data->Iex.RdTmp.tmp;
 					if (tmp <= MAX_TMP && next_irsb_addr == tmps[tmp]) {
 						lr_store_pc = 1;
+					} else {
+						lr_store_pc = 0;
 					}
 				}
 				break;
--- a/pyvex_c/postprocess.c
+++ b/pyvex_c/postprocess.c
@@ -44,8 +44,17 @@ void arm_post_processor_determine_calls(
 	}
 
 	// Emulated CPU context
-	Addr tmps[MAX_TMP + 1] = { DUMMY };
-	Addr regs[MAX_REG_OFFSET + 1] = { DUMMY };
+	Addr tmps[MAX_TMP + 1];
+	Addr regs[MAX_REG_OFFSET + 1];
+
+    // Initialize context
+    for (i = 0; i <= MAX_TMP; ++i) {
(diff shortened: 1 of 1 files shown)


ltfish:
Questions regarding this patch:

- `i` is not initialized. Did you miss an `int i;` at the beginning of this function?
- In the assembly snippet you provided, where is that LR used, and how is it used after the BEQ jump? Is it used as a jump target, or a return target?

danse-macabre:
@ltfish 

> i is not declared. Did you miss an int i; at the beginning of this function?

Yep, I missed that. And now I'm curious, why that didn't break everything...

> In the assembly snippet you provided, where is that LR used, and how is it used after the BEQ jump? Is it used as a jump target, or a return target?

The CFG of the function is pretty complex and I can't really trace how the LR is used after that block, but I believe it is used as a jump target. Nevertheless, I think it doesn't matter since `arm_post_processor_determine_calls()` doesn't look what's happening outside the given block, and on the block-scope level the LR will definitely not point to the next block address.

ltfish:
> The CFG of the function is pretty complicated and I can't really trace how the LR is used after that block, but I believe it is used as a jump target.

If that is the case, I think as of now, angr's CFGFast will be unable to determine the target of the indirect jump :/

danse-macabre:
@ltfish Thanks for the reply!

> IMO the correct fix would be properly handling PUT(LR) = xxx when xxx is neither PC nor a constant value by saving the value of LR.

Could you please clarify what do you mean by "saving the value of LR"?

> Your correct fix seems like a no-op to me as lr_store_pc starts off as a 0 anyway.

Let me disagree with you. The second commit ensures that `lr_store_pc` is set to 1 only if the last value written to LR was the address of the next block. Consider the following example:

```
1000	sub_1000:
1004			ldr	lr, =(sub_1014)
1008			mov r0, lr
100c			ldr	lr, =(sub_1100)
1010			bx 	lr
1014	sub_1014:
1014			...
...			...
1100	sub_1100:
1100			...
...			...
```

Without the fix `lr_store_pc` will be set to 1 while processing statements for instruction at 0x1004, and it will retain its value until the end of the block despite the fact that it is overwritten at 0x100c. The example is from the top of my head, but who knows what code lurks in those "real-world" binaries.

PR #180, opened 2019-02-05
Fixes pathname issue with "()" in Path.

adamdoupe:
Despite the insane sq function used to handle paths with spaces, if you have a path with parenthesis (which Dropbox does by default when you have multiple Dropbox accounts) pyvex will not compile. 

Even though I'm using symbolic links so that there is no space or parenthesis, pyvex blows up (something is expanding to the full path name).

This usage of the `call sq` is not needed, because it's used as part of the path and can be enclosed in double quotes.


rhelmot:
This does in fact seem to work still on my dumbass setup with a space in the path. Thanks!

PR #182, opened 2019-02-28
Support optimization in consumers by providing continuous natural number tags

ltfish:


ltfish:
The failed CLE test cases should be unrelated to this PR.

PR #188, opened 2019-03-31
adjust sample code in readme for python3

jasperla:
This adjusts the sample code in the readme to be valid Python 3 syntax.

It doesn't fully work however as it results in the following for me:

```
Traceback (most recent call last):
  File "/tmp/readme.py", line 50, in <module>
    print(irsb.tyenv.types[0])
IndexError: list index out of range
```

But I cannot check whether this was already the case when pyvex was still Python 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,16 +30,16 @@ import pyvex
 import archinfo
 
 # translate an AMD64 basic block (of nops) at 0x400400 into VEX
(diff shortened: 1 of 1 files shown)


rhelmot:
The sample is failing because the block is all nops, and thus has no temps. No idea if it ever worked in the past. Thank you so much!

PR #198, opened 2019-10-14
Type Annotations for pyvex

fmagin:
Initial effort to add type annotation to the angr project. This should have no effect on runtime behavior, but makes developing pyvex and anything that uses it easier. At least for the project I am currently working on.

This can be merged incrementally, for now this is WIP until I have at least covered the annotations that I need to statically type another poject. PEP 561 explicitly allows a package that is only partially typed.

### DIFF
--- a/pyvex/__init__.py
+++ b/pyvex/__init__.py
@@ -14,6 +14,7 @@ import sys
 import cffi
 import pkg_resources
 RegisterOffset = NewType('RegisterOffset', int)
+TmpVar = NewType('TmpVar', int)
 from .vex_ffi import ffi_str as _ffi_str
 ffi = cffi.FFI()
 
--- a/pyvex/__init__.py
+++ b/pyvex/__init__.py
@@ -2,6 +2,7 @@
 PyVEX provides an interface that translates binary code into the VEX intermediate represenation (IR).
 For an introduction to VEX, take a look here: https://docs.angr.io/advanced-topics/ir
 """
+from typing import NewType, Any
 
 __version__ = (8, 20, 1, 7)
 
@@ -12,10 +13,9 @@ import os
 import sys
 import cffi
 import pkg_resources
-
+RegisterOffset = NewType('RegisterOffset', int)
 from .vex_ffi import ffi_str as _ffi_str
 ffi = cffi.FFI()
-
 
 import logging
 logging.getLogger("pyvex").addHandler(logging.NullHandler())
@@ -40,7 +40,7 @@ def _find_c_lib():
     dir(lib)
     return lib
 
-pvc = _find_c_lib()
+pvc = _find_c_lib() # type: Any # This should be properly typed, but this seems non trivial
 
 # pylint: disable=wildcard-import
 from .enums import *
--- a/pyvex/__init__.py
+++ b/pyvex/__init__.py
@@ -13,9 +13,6 @@ import os
 import sys
 import cffi
 import pkg_resources
-RegisterOffset = NewType('RegisterOffset', int)
-RegisterName = NewType('RegisterName', str)
-TmpVar = NewType('TmpVar', int)
 from .vex_ffi import ffi_str as _ffi_str
 ffi = cffi.FFI()
 
--- a/pyvex/__init__.py
+++ b/pyvex/__init__.py
@@ -14,6 +14,7 @@ import sys
 import cffi
 import pkg_resources
 RegisterOffset = NewType('RegisterOffset', int)
+RegisterName = NewType('RegisterName', str)
 TmpVar = NewType('TmpVar', int)
 from .vex_ffi import ffi_str as _ffi_str
 ffi = cffi.FFI()
--- a/pyvex/block.py
+++ b/pyvex/block.py
@@ -1,6 +1,10 @@
 import copy
 import sys
 import itertools
+from typing import List, Optional
+
+from pyvex.expr import IRExpr
+from pyvex.stmt import IRStmt
 
 from . import VEXObject
 from . import expr, stmt
@@ -83,10 +87,10 @@ class IRSB(VEXObject):
         self.addr = mem_addr
         self.arch = arch
 
-        self.statements = []
-        self.next = None
+        self.statements = [] # type: List[IRStmt]
+        self.next = None # type: Optional[IRExpr]
         self._tyenv = None
-        self.jumpkind = None
+        self.jumpkind = None # type: Optional[str]
         self._direct_next = None
         self._size = None
         self._instructions = None
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -3,7 +3,8 @@ import logging
 from typing import List, Optional
 
 
-from . import VEXObject, RegisterOffset, TmpVar
+from . import VEXObject
+from archinfo import RegisterOffset, TmpVar
 from .enums import IRCallee, IRRegArray, get_int_from_enum, get_enum_from_int
 from .const import get_type_size, U8, U16, U32, U64
 
--- a/pyvex/expr.py
+++ b/pyvex/expr.py
@@ -3,14 +3,12 @@ import logging
 from typing import List, Optional, NewType
 
 
-from . import VEXObject
+from . import VEXObject, RegisterOffset, TmpVar
 from .enums import IRCallee, IRRegArray, get_int_from_enum, get_enum_from_int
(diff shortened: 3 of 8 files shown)


pyvex/__init__.py
```diff
@@ -40,7 +40,7 @@ def _find_c_lib():
     dir(lib)
     return lib
 
-pvc = _find_c_lib()
+pvc = _find_c_lib() # type: Any # This should be properly typed, but this seems non trivial
```
twizmwazin: Probably leave this un-annotated if not planning to properly type it out?

pyvex/block.py
```diff
@@ -83,10 +87,10 @@ def __init__(self, data, mem_addr, arch, max_inst=None, max_bytes=None,
         self.addr = mem_addr
         self.arch = arch
 
-        self.statements = []
-        self.next = None
+        self.statements = [] # type: List[IRStmt]
```
twizmwazin: This syntax is not necessary any more, post python 3.5 iirc. You can define instance variables at the class level, along with their type. Example:

```py
class Person:
    name: str

    def __init__(self, name: str):
        self.name = name

```

pyvex/expr.py
```diff
@@ -15,14 +20,14 @@ class IRExpr(VEXObject):
...
-    tag = None
+    tag = None # type: Optional[str]
     tag_int = 0  # set automatically at bottom of file
 
     def pp(self):
         print(self.__str__())
 
     @property
-    def child_expressions(self):
+    def child_expressions(self) -> List['IRExpr']:
```
twizmwazin: I'm not sure that this is valid, shouldn't this be some IRExpr type rather than the literal string "IRExpr" ?

rhelmot: It is valid.

pyvex/block.py
rhelmot: That syntax doesn't kick in until py 3.6. It is an error on 3.5 and we still need to support 3.5.

pyvex/__init__.py
rhelmot: https://github.com/angr/pyvex/blob/master/pyvex_c/pyvex.def ;)

pyvex/expr.py
fmagin: This is a [forward reference](https://www.python.org/dev/peps/pep-0484/#forward-references). This is required in various places because `def child_expressions(self) -> List[IRExpr]` would require the type to be available when this definition is parsed. The import only happens later (or not at all?), so this would be a syntax error. This is needed in various places and still properly type checks

fmagin:
> This is awesome, I'd like to see type hints throughout angr.

This is the goal, Pyvex is just what I needed first for the actual project I am typing at work.

pyvex/block.py
fmagin: I am still targeting 3.5+ and have my PyCharm configured as such that breaking compatibility is an error.
There are a few pitfalls like that and as far as I saw automatic conversion is an option anyway, so I would suggest waiting for at least pypy 3.6 to be stable before dropping 3.5 support.

fmagin:
Another goal of this before (hopefully still in this PR) will be to integrate at least the basic type checking into the CI. Meaning that any definitive type errors will be caught and fail the CI. In the future the only thing required would then be to change the mypy config to be stricter.

pyvex/expr.py
twizmwazin: Ahh, gotcha. TIL, thanks for the explanation!

twizmwazin:
> Another goal of this before (hopefully still in this PR) will be to integrate at least the basic type checking into the CI. Meaning that any definitive type errors will be caught and fail the CI. In the future the only thing required would then be to change the mypy config to be stricter.

Yes! We can have this added to CI. I'm (intermittently) working on expanding our CI to do nightly builds, so we can roll mypy checking in with that. After yesterday's new pypy release, we may be re-targeting to python 3.6 in a future release, which would give us more options for type hints as well.

fmagin:
3.6 doesn't actually bring that many cool typing features (a lot of cool things are 3.7 and 3.8) so I see no need to rush. The non comment annotations are nice, but nothing I'd feel like immediately breaking 3.5 compatibility for.

For now the default mypy output has way too many errors to integrate into the CI, so we have to either configure mypy more leniently or to fix the 600-1000 errors. The latter might be easer than it sounds, I think I reduced it by a few hundreds just with https://github.com/angr/angr/pull/1752/commits/19e6595f842f177329af4f22cf9b43d876868030

fmagin:
I correct my previous statement: 3.6 supports [PEP 526: Variable Annotations](https://www.python.org/dev/peps/pep-0526/) which seems needed to annotate the state plugins of a [SimState](https://github.com/fmagin/angr/blob/7b310aa91b2792dfec6f10a392982d743a82cb80/angr/sim_state.py#L60-L64). Without this any access like `state.solver` or `state.regs` will need a cast because the type checker can just infer the type as `SimStatePlugin`

This has decent advantages for development. I found various bugs in angr-cli, for example code that assumed a specific heap implementation (`SimHeapBrk`) that has a certain attribute (`heap_location`). In general it allows tab completion on the actual plugin and any type inference from this.

The question is: Do we want to break Python 3.5 compatibility over this, ignore typing SimStatePlugins (and probably other code with similar patterns, knowledge base maybe) or find some workaround? An example for the latter would be initializing the variables with something (`None`) to allow the older type annotations and changing the plugin loading/accessing code to treat `None` the same as the attribute not existing. I'd rather avoid this, and because everyone else seemed eager to drop 3.5 now that pypy is on 3.6 I would not block and rather encourage this.

fmagin:
Because various parts of angr would greatly benefit from even more advanced typing features (Named Dicts with Literal Types to specify valid configuration options which might allow auto completing something like `load_options={'auto_lo<tab>` to `load_options={'auto_load_libs':`
I am now planning to move all annotations to `.pyi`  files (`stubgen` supports this) which should allow type checking with higher python versions while keeping the code base compatible with 3.5. I would still contribute them to main repos (instead of a `-stub` package) and place them next to actual source files. I think a separate folder is an option if this is preferred. But that would make it less obvious that they exist and might be harder to setup with IDEs

rhelmot:
I worry about the use of stubs. The hard part with them is making sure they stay in sync with the code. Can CI be configured to fail if you change a function signature but not the stub signature?

fmagin:
I would assume that it

fmagin:
https://github.com/python/mypy/issues/5028 talks about this, there is a mildly hacky solution that merges the stub files into the source files (which would error when the signature mismatches). I do miss mypy complaining about type errors in the actual code because it just checks the pyi files.

Ultimately I am not sure what is preferable.  We could drop python 3.5 support and require 3.6+ to use class variable annotations (which are required to type annotate the SimStatePlugins which is essential IMO), and then use `typing_extensions` to use the 3.7/3.8 features (which would be cool, but not essential I think. The thought of never having to type out "auto_load_libs" again is tempting though.)

fmagin:
I skimmed some of the linked issues and it seems that the retype solution is fairly painful and retype itself not quite sufficient feature-wise. Looking at the mess of import errors in the stubgen generated stubs I am not looking forward to dealing with those either. So I feel like it ultimately comes down to:
Are type annotations that are remotely useful[0] worth dropping python 3.5 support over? 
We should make sure that it works with pypy 3.6, but I don't see any reason to cater to older Ubuntu LTS releases except people with old distros asking questions in the Slack why it doesn't work, but I think I'd rather deal with those then lack of type annotations. 


[0]  which IMO they wouldn't when any SimStatePlugin access looses the type information

fmagin:
The only test failing is the one discussed here https://github.com/angr/vex/pull/26#issuecomment-577648456

pyvex/block.py
twizmwazin: Before merging this, do we want to reconsider since we've now moved to targeting 3.6?

fmagin: At this point I would just really like this to merged asap so I don't have to rebase all the time and use my own angr fork for other things I am working on that require type annotations.

 I am not sure if this makes any functional difference or has any advantage. https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations sounds like it might allow to declare that a variable must be initialized after `__init__` while allowing it to be set to `None` at some point during `__init__`. I think I had this issue in some places, so if this solves this I am definitely in favor of changing that.

PR #199, opened 2019-11-05
Start aarch64 spotter, add MSR, MRS and SYSL support

datadancer:
Fix [angr#1783](https://github.com/angr/angr/issues/1783)

### DIFF
--- a/pyvex/lifting/gym/__init__.py
+++ b/pyvex/lifting/gym/__init__.py
@@ -1,4 +1,5 @@
 from .arm_spotter import ARMSpotter
 from .x86_spotter import X86Spotter
+from .aarch64_spotter import AARCH64Spotter
 
-__all__ = ('ARMSpotter', 'X86Spotter')
+__all__ = ('ARMSpotter', 'X86Spotter', 'AARCH64Spotter')
--- a/pyvex/lifting/gym/aarch64_spotter.py
+++ b/pyvex/lifting/gym/aarch64_spotter.py
@@ -1,11 +1,14 @@
+import logging
+
 from ..util.lifter_helper import GymratLifter
 from ..util.instr_helper import Instruction
 from .. import register
-import logging
 
 l = logging.getLogger(__name__)
 
 class Aarch64Instruction(Instruction): # pylint: disable=abstract-method
+    # NOTE: WARNING: There is no MRS, MSR, SYSL in VEX's ARM implementation
+    # You must use straight nasty hacks instead.
     pass
 
 class Instruction_SYSL(Aarch64Instruction):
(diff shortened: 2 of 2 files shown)


datadancer:
Fix [angr#1783](https://github.com/angr/angr/issues/1783)

rhelmot:
The way logging formatter arguments is supposed to work in python is that you say `l.debug("fmt %s %s", foo, bar)` and the formatting is done lazily, conditional on the message actually being output somewhere. That way, it saves the cycles that would be spent on string rendering if they're not necessary. Therefore, doing eager formatting with the % operator in a logging statement is a lint error. Your previous commit was incorrect because the logging arguments are supposed to be provided as varargs, not an iterable single argument.

The docs error needs to be solved with a commit to angr-doc, which I will do when this is merged. The other error is spurious. Fix the lint and I'll hit merge.

datadancer:
All tests OK except for the api-doc case. Please merge this branch by adding the following line to file api-doc/source/pyvex.rst
`.. automodule:: pyvex.lifting.gym.aarch64_spotter`

datadancer:
HELP! 
I can not find the test case test_boolector_solving in the the angry-dev project. Where do I find it and test it by hand? Thanks a lot.
```
+ nose2 -v -s ./src/rex/tests -c nose2.cfg --log-level 100 test_rex.test_boolector_solving test_explore.test_write_what_where_shadowstack
qemu: uncaught target signal 15 (Terminated) - core dumped [0814042a]
test_rex.transplant_class.<locals>.C (test_boolector_solving) ... qemu: uncaught target signal 15 (Terminated) - core dumped [080486d3]
ok
```

twizmwazin:
Test binaries should be in the `binaries` repo, which angr-dev should have checked out in the same directory as angr, pyvex, etc.

datadancer:
Thanks for your advice. However, I can not find such file in the `binaries` repo using the following `find` and `grep` commands.
```
bing@YT:/workspace/angr/angr-dev$ find binaries/ | grep -i boolector
bing@YT:/workspace/angr/angr-dev$ grep -i boolector -rn binaries/
```

rhelmot:
Kevin, they weren't asking for a test binary, they were asking where the testcase lives. it's [here](https://github.com/shellphish/rex/blob/master/tests/test_rex.py#L143)

However, the snippet you posted is actually a log of the test passing. It is a test of an automatic exploitation framework, so the segfault is the exploit in action. You can see from the `ok` that the test passes. You can see from [azure](https://dev.azure.com/angr/pyvex/_build/results?buildId=944&view=ms.vss-test-web.build-test-results-tab) that the only thing failing is the api-doc coverage test, which I will do for you.

Thanks!

PR #201, opened 2020-01-22
Add @rpath to the library install name on macOS

saagarjha:
This makes it possible for angr to link against libpyvex.dylib more easily on macOS. For context on why this change would be useful, see https://github.com/angr/angr/pull/1932.


rhelmot:
Interesting! I know absolutely nothing about macos library linking, so I am going to accept this blindly. It would be nice if we could stop having such serious issues with macs.

PR #204, opened 2020-02-29
VEX: Support disabling cross-instruction optimizations.

ltfish:
Sync https://github.com/angr/vex/pull/35.

### DIFF
--- a/pyvex/lifting/libvex.py
+++ b/pyvex/lifting/libvex.py
@@ -27,6 +27,8 @@ class VexRegisterUpdates:
 
 class LibVEXLifter(Lifter):
 
+    __slots__ = ()
+
     REQUIRE_DATA_C = True
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


pyvex/lifting/libvex.py
```diff
@@ -16,12 +16,25 @@
...
 class LibVEXLifter(Lifter):
 
-    __slots__ = ()
+    __slots__ = ('_px_control', )
 
     REQUIRE_DATA_C = True
 
+    def __init__(self, arch, addr):
+        super().__init__(arch, addr)
+        self._px_control = ffi.new("VexRegisterUpdates *")
```
rhelmot: We don't ever have any need to do anything other than pass this value in from python, so why are we passing a pointer here? We can just have pyvex.c do whatever referencing needs to be done.

PR #212, opened 2020-05-10
Fixes VS2019 Build Error

bannsec:
When attempting to build pyvex with VS2019, the compilation process was throwing an error about not having `pub\libvex_guest_offsets.h`. As it turns out, vex doesn't by default generate that file. The command to do so is in the makefile but needs to be run. This is why build is failing on msvc.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ def _build_vex():
     e['MULTIARCH'] = '1'
     e['DEBUG'] = '1'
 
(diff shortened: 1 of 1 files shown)


rhelmot:
I would prefer this to be fixed by editing the makefile to make the header (denoted by GEN_HEADERS) a dependency of the objects that need it to be built.

bannsec:
kk

PR #218, opened 2020-06-16
Implemented tMRS/tMSR spotters

degrigis:
This is the spotter for the tMRS instruction.
Before merging, 2 quick questions:

1. Maybe there is a better way to refer to the SP in that `self.get(13, Type.int_32)` instead of hardcoding the index in the register list?

2. As for now, I've implemented only the minimum checks I need for my analysis, do we want to start to organize `archinfo` with the information regarding ALL the special registers we are missing or it is too early for this? (See https://github.com/aquynh/capstone/blob/45bec1a691e455b864f7e4d394711a467e5493dc/arch/ARM/ARMInstPrinter.c#L1654)

### DIFF
--- a/pyvex/__init__.py
+++ b/pyvex/__init__.py
@@ -4,7 +4,7 @@ For an introduction to VEX, take a look here: https://docs.angr.io/advanced-topi
 """
 from typing import NewType, Any
 
-__version__ = (8, 20, 7, 27)
+__version__ = (9, 0, "gitrolling")
 
 if bytes is str:
     raise Exception("This module is designed for python 3 only. Please install an older version to use python 2.")
--- a/pyvex/lifting/gym/arm_spotter.py
+++ b/pyvex/lifting/gym/arm_spotter.py
@@ -263,7 +263,7 @@ class Instruction_tCPSID(ThumbInstruction):
         l.debug("[thumb] Ignoring CPS instruction at %#x.", self.addr)
 
 class Instruction_tMSR(ThumbInstruction):
-    name = 'MSR'
+    name = 'tMSR'
     bin_format = '10x0mmmmxxxxxxxx11110011100Rrrrr'
 
     def compute_result(self): # pylint: disable=arguments-differ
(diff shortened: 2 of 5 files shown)


degrigis:
1. Fixed
2. ?

degrigis:
I've implemented the missing test for tMRS, however, the lifting of the instruction happens only if I use opt_level=3 when building the IRSB, is it ok tho? @subwire @ltfish

ltfish:
> Maybe there is a better way to refer to the SP in that self.get(13, Type.int_32) instead of hardcoding the index in the register list?

Is archinfo available here? You can use `arch.sp_offset` if it is available.

> As for now, I've implemented only the minimum checks I need for my analysis, do we want to start to organize archinfo with the information regarding ALL the special registers we are missing or it is too early for this?

I think that is a good idea, but it can happen as a separate wave of PRs.

> I've implemented the missing test for tMSR, however, the lifting of the instruction happens only if I use opt_level=3 when building the IRSB, is it ok tho?

Question: Why is this?

degrigis:
> > Maybe there is a better way to refer to the SP in that self.get(13, Type.int_32) instead of hardcoding the index in the register list?
> 
> Is archinfo available here? You can use `arch.sp_offset` if it is available.
> 

Technically no. 
Don't we want to avoid inter-modules imports? It seems that nowhere in `pyvex`  there is an import for archinfo.

> > As for now, I've implemented only the minimum checks I need for my analysis, do we want to start to organize archinfo with the information regarding ALL the special registers we are missing or it is too early for this?
> 
> I think that is a good idea, but it can happen as a separate wave of PRs.
> 

Ok, agree! 
> > I've implemented the missing test for tMSR, however, the lifting of the instruction happens only if I use opt_level=3 when building the IRSB, is it ok tho?
> 
> Question: Why is this?

Very good question, I have not investigated, do you want me to do that to close this PR?

degrigis:
I've also just discovered that accessing the `primask` in that way is not safe, not all the ARM architectures have it apparently.

ltfish:
Sorry for the delayed response. Is this PR ready?

degrigis:
Yep, I would merge this and fix possible extra issues in the future.

degrigis:
Is the issue with the `colorguard` test known?  @ltfish

rhelmot:
As far as I can tell, this is not a normal test failure, it doesn't normally fail on other branches, but it shouldn't be related to this PR at all... I think it should be safe to ignore.

degrigis:
Yeah because technically I've just changed the spotter for tMRS/tMSR. I'll quickly investigate and report back!

PR #219, opened 2020-07-17
Remove x86 spotter; endbr is now in libvex

rhelmot:
sync https://github.com/angr/vex/pull/38


rhelmot:
CI seems... broken. I am going to merge this and if problems come up we'll deal with them.

PR #232, opened 2021-05-28
added spotter for tMRS/tMSR

degrigis:
See https://github.com/angr/pyvex/pull/218


degrigis:
Ok, we should be good now @ltfish @rhelmot
(PR https://github.com/angr/pyvex/pull/218 was failing a `colorguard` test for some unknown reason)

ltfish:
Merged. Thanks!

PR #253, opened 2022-01-17
Fix unable to compile pyvex on macOS bug.

adamdoupe:
Issue was commit 80c9beba7f450c0ae1a1997eeb3a291ebed40490 moved various functions to pyvex_internal.h from pyvex.h. postprocess.c uses functions from pyvex_internal.h but was only including pyvex.h


ltfish:
Thanks!

PR #254, opened 2022-02-16
minor type annot

ulugbekna:
- type annotate `irsb_c` in `VexValue` constructor
- type annotate `rdt` to `RdTmp | Const`


rhelmot:
Thanks!

PR #255, opened 2022-02-17
add shift arith right vex

ulugbekna:
Adds support for arithmetic shift right for Vex and some comments.


rhelmot:
Glad you figured this out :) Thanks!

PR #271, opened 2022-04-05
Remove pycparser dependency

twizmwazin:


ltfish:
Why did we have this dependency to begin with…

twizmwazin:
Seems to date back to [this](https://github.com/angr/pyvex/commit/21c0c6f3ed0a243b5e8352d06a43e3030dd0fe1e) commit, when it was added to `requirements.txt`. I checked out that version of the code and found no other references to pycparser at that time.

rhelmot:
cffi and pycparser have some sort of interdependent relationship. In the past I believe cffi didn't require a new enough version of pycparser, so we got a lot of errors reported where peoples' pycparsers were too old.

PR #276, opened 2022-05-19
Fix uncast bit index

rhelmot:
Closes #185


rhelmot:
This also breaks CI (there are spurious failures right now but there's a new pyvex failure which seems to matter) so I'm going to close this and let the original PR go stale.

PR #280, opened 2022-06-23
pyvex.c: detect and handle 32-bit powerpc, at least on NetBSD.

he32:
This lets this package build on NetBSD/macppc.


ltfish:
Are you trying to run PyVEX on a PPC32 host?

ltfish:
I'm sure there is at least one bug in PyVEX that prevents it from working correctly on non-64-bit platforms. Good luck!

PR #308, opened 2023-03-22
plugins 2: no more plugins

rhelmot:
We begin by discarding all previously held notions of what is good and just. The things that are good and just are the things that work. We must not have the dependency cycle; this is bad. Pyvex cannot be refactored to not use arch, this is bad. Making archinfo not depend on pyvex is a fool's errand, this is bad. Therefore, pyvex must put its finger to its lips and claim, "I do not depend on archinfo".

It's still not good. But it does work.


twizmwazin:
Tests are still using archinfo, fixing those up then plan to merge

PR #324, opened 2023-06-05
Log at "debug" level instead of "info"

paulkermann:
We shouldn't log this kind of information at the "info" logging level. "debug" level will be more appropriate, like the other logs done in this function.


twizmwazin:
Thanks!

PR #352, opened 2023-09-22
OSSFuzz Integration

capuanob:
Hello!

I contributed to angr for my master's thesis and, having gotten deeper into fuzzing, thought to integrate pyvex into OSS Fuzz!

This PR introduces the source for the fuzzer and a YAML config to fuzz each PR for a set duration. I would greatly appreciate it being merged in! 

I will continue monitoring OSS-Fuzz for bugs and fix any I can in future PRs. Yan also has access to the OSS Fuzz dashboard to review any security-relevant crashes that may come up.

Please let me know if anything else is needed

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,9 @@
-name: CIFuzz
-on: [pull_request]
+name: OSS-Fuzz
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 permissions: {}
 jobs:
  Fuzzing:
--- a/fuzzing/build.sh
+++ b/fuzzing/build.sh
@@ -1,0 +1,33 @@
+#!/bin/bash -eu
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# Since pyvex requires a specific developer build of archinfo, install it from source
+cd "$SRC"/archinfo
+pip3 install .
+
+cd "$SRC"/pyvex
+pip3 install .[testing]
+
+# Generate a simple binary for the corpus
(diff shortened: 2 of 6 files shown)


ltfish:
Hi @capuanob , thanks for your contribution!

I wonder why you decide to (use OSSFuzz to) fuzz pyvex instead of libVEX in Valgrind. Thoughts?

twizmwazin:
It is not obvious to me how the pipeline description translates into the steps I see being done in the log. Is this process stored elsewhere? If we make changes in the future that would require tweaking the dependencies and install process, how would we go about keeping it working?

capuanob:
The remainder of the configuration can be seen on the OSSFuzz project in
the projects/pyvex directory. Should there be a breaking change to the
build process, an alert would go out that the nightly OSS-Fuzz building has
failed for N (I believe three?) successive days. It would then require a PR
to update the build.sh in that project to reflect the new install process.

capuanob:
To note, I will also get notifications and will try to maintain successful
fuzzing on the projects that are tied to my email.

twizmwazin:
Gotcha. It seems the files in `fuzzing` are duplicated with the code in the oss-fuzz repo. Is that intentional? Which ones are actually being used?

capuanob:
Currently, the ones in the OSS-Fuzz project. Once this gets merged, I will
switch it over. Didn't want to have downtime in fuzzing, but hosting it
here is preferred by Google and will allow for easier updates (more
harnessing or changes to existing fuzzers) than requiring PRs to OSS-Fuzz.

twizmwazin:
In that case, would it be possible to also move the build process to this repo? And if someone makes a change that causes the harness to break, would the CI fail so we can fix it?

capuanob:
As I understand it, the build.sh, Dockerfile, and project.Yaml must reside
in oss-fuzz.

I could think it would be possible to have build.sh call another bash
script for build instructions, but it would take some experimentation.

If possible, I’d appreciate if we could merge this in first and then I can
look into that once it’s working again

Thank You, Bailey Capuano

capuanob:
> In that case, would it be possible to also move the build process to this repo? And if someone makes a change that causes the harness to break, would the CI fail so we can fix it?

As to the second question,

Should the harness be changed in a way that makes it exit instantly or unable to be called, the CI will fail when building as it tests to ensure that it can fuzz for a few iterations

capuanob:
@twizmwazin I've looked into your question a bit more. The Dockerfile and project.yaml must be hosted in the OSS-Fuzz project. However, we could move the build.sh into the pyvex source. I'll push that up momentarily.

capuanob:
@ltfish @twizmwazin Where does this PR stand? Is anything else needed for approval?

Thank you for your time in reviewing!

PR #367, opened 2023-11-25
Fix gymrat disasm

xxr0ss:
The `disassemble()` doesn't work because `GymratLifter._lift()` doesn't receive its args, which are not passed through `Lifter.lift()`

```python
class GymratLifter(Lifter):
    ...
    def _lift(self, disassemble=False, dump_irsb=False):
```

and disasm result isn't returned from `Lifter.lift()`, which returns `self.irsb` instead of `self._lift()`:
```python
class Lifter:
    ...
    def lift(
        self,
        data,
        bytes_offset=None,
        max_bytes=None,
        max_inst=None,
        opt_level=1,
        traceflags=None,
        allow_arch_optimizations=None,
        strict_block_end=None,
        skip_stmts=False,
        collect_data_refs=False,
        cross_insn_opt=True,
        load_from_ro_regions=False,
    )
        ...
        self._lift()
        return self.irsb
```

I think `lift()` should return `IRSB` only because of its sematic, however `GymratLifter` (as well as many other code not listed here) also uses `_lift()` to return disassembly. Well, let's allow that.

### DIFF
--- a/pyvex/lifting/util/lifter_helper.py
+++ b/pyvex/lifting/util/lifter_helper.py
@@ -38,8 +38,6 @@ class GymratLifter(Lifter):
         "bitstrm",
         "errors",
         "thedata",
-        "disasm",
-        "dump_irsb",
     )
 
     REQUIRE_DATA_PY = True
@@ -50,8 +48,6 @@ class GymratLifter(Lifter):
         self.bitstrm = None
         self.errors = None
         self.thedata = None
-        self.disasm = False
-        self.dump_irsb = False
 
     def create_bitstrm(self):
         self.bitstrm = bitstring.ConstBitStream(bytes=self.thedata)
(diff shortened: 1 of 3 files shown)


twizmwazin:
I'm not a fan of this solution, instead I would rather make the gymrat options be explicit kwargs on `lift()` instead and have them set as instance variables to be optionally read in `_lift`, just like the many libvex-specific args.

xxr0ss:
@twizmwazin I agree. Let's just override the `lift()`!

xxr0ss:
@twizmwazin thanks for the detailed explanation!

twizmwazin (approved):

pyvex/lifting/util/lifter_helper.py
```diff
@@ -136,11 +138,13 @@ def pp_disas(self):
...
         print(disasstr)
 
     def error(self):
         return self.errors
 
     def disassemble(self):
-        return self.lift(disassemble=True)
+        if self.disassembly is None:
+            self.lift(self.data, disasm=True)
+        return self.disassembly
```
xxr0ss, line 150: Introduced a new variable `self.disassembly` here to save the disasm result, since `Lifter.lift()` returns irsb, when disassembly is returned from `GymratLifter`'s overriden `_lifter`, it won't be passed back to the caller.

PR #474, opened 2025-06-25
Switch build system to use scikit-build-core and cmake

twizmwazin:


rhelmot:
why scikit? do they just have a good cmake integration?

twizmwazin:
yep. It seems to be the simplest way to use cmake

PR #487, opened 2025-07-31
Fix: Allow multiple users to load the cache.

### LINKED ISSUE #486: Running Angr on Linux with multiple users results in clashes
### Description

This code (_parse_ffi_str):
https://github.com/angr/pyvex/blob/master/pyvex/native.py#L16
Creates a cache file such as this:
/tmp/pyvex_ffi_parser_cache.27654ee7f970352e83b36ba81f53e1b2
But the file is read/write only to the owner, and not readable by others.  The next user to run Angr will get an exception raised, because the file is known to exist, and _parse_ffi_str attempts to open it, but can't, and an exception is raised.  This basically prevents two different users on a Linux system from importing pyvex.

### Steps to reproduce the bug

On a Linux system, with two users, run any Angr script that imports pyvex as user 'A'.  Then run the same thing as user 'B'.  The user 'B' attempt should fail, because of an exception trying to open the FFI cache file.

### Environment

_No response_

### Additional context

_No response_

ltfish:
Closes #486.

### DIFF
--- a/pyvex/native.py
+++ b/pyvex/native.py
@@ -17,13 +17,20 @@ def _parse_ffi_str():
     hash_ = hashlib.md5(_ffi_str.encode("utf-8")).hexdigest()
     cache_location = os.path.join(tempfile.gettempdir(), f"pyvex_ffi_parser_cache.{hash_}")
 
+    cache_loaded = False
     if os.path.isfile(cache_location):
         # load the cache
-        with open(cache_location, "rb") as f:
-            cache = pickle.loads(f.read())
-        ffi._parser._declarations = cache["_declarations"]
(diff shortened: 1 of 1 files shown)


rhelmot:
I don't know if this is a worthwhile concern, but this would allow fairly trivial privesc from one local user to another which is using angr. Maybe use $XDG_RUNTIME_DIR if it's set and fallback to /tmp if it's not?

rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/pyvex_487

rhelmot:
sticky bit to the rescue! @ltfish final approval?

PR #495, opened 2025-09-10
Simple test for masking the shift amount in riscv64 rtype shift instructions

Cskorpion:
Test for [#77](https://github.com/angr/vex/pull/77)


rhelmot:
It'll be a little weird to test this in CI because of the vex submodule. If you verify that this passes locally, I'll merge both the PRs and add another commit re-pinning the submodule.

PR #541, opened 2026-05-18
pyvex_c: widen HashHW value slot to ULong (fix 64-bit guest-pointer truncation on 32-bit hosts)

### LINKED ISSUE #539: pyvex IRSB.data_refs truncates 64-bit RIP-relative load addresses to 32 bits on wasm32-emscripten
### Description

On `wasm32-emscripten`, `pyvex.lift(..., collect_data_refs=True)` reports the `data_addr` of a memory `data_ref` as its low 32 bits only when the address is *derived from* (not produced by) an earlier instruction in the same block. The same `pyvex.lift()` call on `x86_64-linux-gnu` produces the full 64-bit address. The IRSB itself (size, jumpkind, instruction count, statement count, statement list) is byte-identical between the two builds; only the `data_refs` differ.

This silently breaks angr's `CFGFast` static analysis on PE images with `ImageBase >= 0x1_00000000` (i.e. essentially all 64-bit Windows binaries):

1. `CFGFast._process_irsb_data_refs` checks `if not self._seg_list.is_occupied(ref.data_addr)` before calling `self._seg_list.occupy(ref.data_addr, ref.data_size, "unknown")`.
2. With the truncated `data_addr`, `is_occupied` either succeeds against the wrong bytes or fails to mark the *correct* bytes as data.
3. In the next pass `CFGFast`'s complete-scan picks up those bytes as if they were code, lifts a one-instruction block from the jump-table data, and creates a spurious function entry.
4. The spurious function entry perturbs the `SpillingFunctionDict` LRU state, so `drop_bad_functions` decides differently for unrelated nearby functions, which cascades into off-by-a-few-bytes function-start divergence elsewhere in the binary.

Observed on `DPInst.exe` (1 MB Windows installer): without any workaround, wasm-angr emits 7219 functions vs. native's 5360 (≈35% inflation). With a downstream rebase workaround applied selectively to `data_refs` whose values clearly look truncated, wasm-angr converges to within ±4 functions of native, confirming the cascade but also indicating at least one related bug elsewhere in the same code path (a derived address in some blocks remains truncated even when re-rebased, suggesting a second affected value-flow site).

### Steps to reproduce the bug

Run the script below that will show an end-to-end reproduction issue.

The PoC lifts a 25-byte x86-64 block (5 instructions, taken verbatim from `DPInst.exe` in the Windows DDK at VA `0x10001b2e4`):

```
  0x10001b2e4:  lea  rdi, [rip - 0x14fc7]        ; rdi = 0x100006324
  0x10001b2eb:  mov  ebp, 0x14
  0x10001b2f0:  lea  rdx, [rsp + 0x20]
  0x10001b2f5:  mov  ecx, dword ptr [rdi - 4]    ; load from 0x100006320
  0x10001b2f8:  call 0x10002e370
```

Expected behavior:
  Native pyvex prints (full `data_refs` list, only relevant rows shown):
```
  ins=0x10001b2e4  data_addr=0x100006324  size=0  type=unknown
  ins=0x10001b2f5  data_addr=0x100006320  size=0  type=unknown
  ins=0x10001b2f5  data_addr=0x100006320  size=4  type=integer
  PASS: data_addr=0x100006320
```

Actual behavior:
  Wasm pyvex (built from the same sdist) prints:
```
  ins=0x10001b2e4  data_addr=0x100006324  size=0  type=unknown
  ins=0x10001b2f5  data_addr=0x6320       size=0  type=unknown
  ins=0x10001b2f5  data_addr=0x6320       size=4  type=integer
  FAIL: data_addr=0x6320, expected 0x100006320
        (got & 0xffffffff) == (expected & 0xffffffff)? True
        high 32 bits dropped: 0x1
```

The single-instruction RIP-relative result from the LEA (`0x100006324`) is correct in both builds. Only the value derived later in the block (`rdi - 4 → 0x100006320`) is truncated.

The script performs the following actions:
1. Clones + activates emsdk 4.0.9 (the version Pyodide 0.29.4 expects)  
2. Creates a Python 3.13 venv and installs pyodide-build==0.29.3 + the 0.29.3 xbuildenv
3. Writes a Pyodide recipe + the 3 minimal patches needed to make pyvex compile for wasm32-emscripten:
    - `0001-wasm32-host-arch.patch`: `VEX_HOST_WORDSIZE=4 `for emscripten
    - `0002-genoffsets-via-emulator.patch`: prepend `${CMAKE_CROSSCOMPILING_EMULATOR}` to the CMake `add_custom_command` that runs the build-time helper    
    - `0003-pyvex-host-arch-wasm.patch`: default `vta.arch_host = VexArchAMD64` (the host backend isn't used on the lifting path)
4. Runs `pyodide build-recipes-no-deps pyvex` → emits a `pyodide_2025_0_wasm32` wheel
5. Sets up a tiny Node + `pyodide` harness that `micropip.installs` the wheel and runs the same `pyvex-poc.py`
6. Prints native vs. wasm output side-by-side

Run `./pyvex-wasm-repro.sh` and get:

```
  NATIVE:  PASS: data_addr=0x100006320
  WASM:    FAIL: data_addr=0x6320, expected 0x100006320
```

Prereqs noted in the header: python3.13, node>=18, cmake, gcc/make, git.  

This should take ~6-10 minutes, where most of the time is spent building pyvex under emscripten.

Of note:
- Pyvex's three build-system patches needed for emscripten are confined to the VEX_HOST_WORDSIZE selection (libvex_basictypes.h), CMake's invocation of the build-time genoffsets helper (cross-compile emulator), and pyvex.c's vta.arch_host selection. None of them touch IR construction, data-ref extraction, or post-lift analysis, the divergence we observe must originate inside the unchanged pyvex C sources.
- The repro is fully reproducible from PyPI sources; no proprietary binary is required (the 25 instruction bytes are inlined in the PoC).

```bash
#!/usr/bin/env bash
#
# End-to-end reproduction for: pyvex 9.2.214 on wasm32-emscripten
# truncates 64-bit RIP-relative data_ref target addresses to their low
# 32 bits.
#
# Builds pyvex from the upstream sdist for wasm32-emscripten (using the
# same toolchain Pyodide 0.29.4 ships: emscripten 4.0.9 + pyodide-build
# 0.29.3) with three small build-system patches, then runs the same
# pyvex.lift() call once on the native python and once inside Pyodide
# under Node. The two outputs diverge on data_addr for a load whose
# address is derived from a previous LEA.
#
# Prerequisites on the host: python3.13, node>=18, cmake, gcc/make,
# git. No Rust required (pyvex 9.2.214 itself has no Rust code ΓÇö the
# Rust toolchain note was about the wider angr build).
#
# Total runtime: ~6-8 min cold (emsdk + xbuildenv download + pyvex
# compile). Re-runs reuse the workdir.
#
# Usage:
#   ./pyvex-wasm-repro.sh [workdir]    # defaults to ./pyvex-wasm-repro

set -euo pipefail

WORKDIR=${1:-$(pwd)/pyvex-wasm-repro}
EMSDK_VERSION=4.0.9
PYODIDE_BUILD_VERSION=0.29.3
PYVEX_VERSION=9.2.214
ARCHINFO_VERSION=9.2.214
PYVEX_SHA256=4e2e1220de8b8cb163d8e500c8eb6eca964184d78aefad392526488cdfb64416

mkdir -p "$WORKDIR"
cd "$WORKDIR"
echo "[repro] workdir: $WORKDIR"

# ---------------------------------------------------------------------------
# 1. emsdk (emscripten toolchain)
# ---------------------------------------------------------------------------
if [ ! -d emsdk ]; then
    echo "[repro] cloning + installing emsdk $EMSDK_VERSION ..."
    git clone --depth=1 https://github.com/emscripten-core/emsdk.git
    ( cd emsdk && ./emsdk install "$EMSDK_VERSION" && ./emsdk activate "$EMSDK_VERSION" )
fi
# shellcheck disable=SC1091
source emsdk/emsdk_env.sh >/dev/null

# ---------------------------------------------------------------------------
# 2. Python venv with pyodide-build
# ---------------------------------------------------------------------------
if [ ! -d venv ]; then
    echo "[repro] creating venv + installing pyodide-build $PYODIDE_BUILD_VERSION ..."
    python3.13 -m venv venv
    # shellcheck disable=SC1091
    source venv/bin/activate
    pip install --quiet --upgrade pip
    pip install --quiet "pyodide-build==$PYODIDE_BUILD_VERSION" "wheel<0.45"
    pyodide xbuildenv install "$PYODIDE_BUILD_VERSION"
fi
# shellcheck disable=SC1091
source venv/bin/activate

# ---------------------------------------------------------------------------
# 3. Recipe + patches for pyvex
# ---------------------------------------------------------------------------
mkdir -p packages/pyvex/patches

cat > packages/pyvex/meta.yaml <<EOF
package:
  name: pyvex
  version: $PYVEX_VERSION
source:
  url: https://files.pythonhosted.org/packages/source/p/pyvex/pyvex-$PYVEX_VERSION.tar.gz
  sha256: $PYVEX_SHA256
  patches:
    - patches/0001-wasm32-host-arch.patch
    - patches/0002-genoffsets-via-emulator.patch
    - patches/0003-pyvex-host-arch-wasm.patch
build:
  type: package
  exports: requested
  cflags: |
    -fPIC
requirements:
  host:
    - cffi
  run:
    - cffi
    - archinfo
    - bitstring
EOF

cat > packages/pyvex/patches/0001-wasm32-host-arch.patch <<'EOF'
Subject: [PATCH] libVEX: recognize wasm32/emscripten as a host architecture

VEX's libvex_basictypes.h selects VEX_HOST_WORDSIZE based on compiler
predefines (__x86_64__, __i386__, ...). None match on wasm32-emscripten,
so the build aborts with "Vex: Fatal: Can't establish the host
architecture". sizeof(void*) is 4 on wasm32.

--- a/vex/pub/libvex_basictypes.h
+++ b/vex/pub/libvex_basictypes.h
@@ -222,6 +222,10 @@
 #elif defined(__tilegx__)
 #   define VEX_HOST_WORDSIZE 8
 #   define VEX_REGPARM(_n) /* */
+
+#elif defined(__EMSCRIPTEN__) || defined(__wasm32__) || defined(__wasm__)
+#   define VEX_HOST_WORDSIZE 4
+#   define VEX_REGPARM(_n) /* */

 #else
 #   error "Vex: Fatal: Can't establish the host architecture"
EOF

cat > packages/pyvex/patches/0002-genoffsets-via-emulator.patch <<'EOF'
Subject: [PATCH] pyvex CMake: run genoffsets via crosscompiling emulator

CMake builds the genoffsets helper as a normal executable and runs it
at configure time. Under cross-compile to wasm32 the "executable" is a
Node.js launcher script; CMake's add_custom_command doesn't prepend
CMAKE_CROSSCOMPILING_EMULATOR automatically, so /bin/sh tries to exec
the .js and fails. Prepend the emulator (empty on native).

--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,7 +105,7 @@ set_target_properties(genoffsets PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BIN

 add_custom_command(
     OUTPUT ${CMAKE_SOURCE_DIR}/vex/pub/libvex_guest_offsets.h
-    COMMAND $<TARGET_FILE:genoffsets> > ${CMAKE_SOURCE_DIR}/vex/pub/libvex_guest_offsets.h
+    COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:genoffsets> > ${CMAKE_SOURCE_DIR}/vex/pub/libvex_guest_offsets.h
     DEPENDS genoffsets
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     COMMENT "Generating pub/libvex_guest_offsets.h"
EOF

cat > packages/pyvex/patches/0003-pyvex-host-arch-wasm.patch <<'EOF'
Subject: [PATCH] pyvex.c: pick a host arch under wasm32/emscripten

pyvex.c selects vta.arch_host based on __amd64__/__i386__/__arm__/...
predefines; none match on wasm32-emscripten and the file fails with
  #error "Unsupported host arch"
vta.arch_host is only consulted when VEX re-encodes IR back to host
instructions ΓÇö guest-IR lifting (the angr static-analysis path) never
touches it ΓÇö so defaulting to VexArchAMD64 is safe.

--- a/pyvex_c/pyvex.c
+++ b/pyvex_c/pyvex.c
@@ -183,6 +183,8 @@
 #  if defined(__riscv_xlen) && (__riscv_xlen == 64)
 	vta.arch_host = VexArchRISCV64;
 #  endif
+#elif defined(__EMSCRIPTEN__) || defined(__wasm32__) || defined(__wasm__)
+	vta.arch_host = VexArchAMD64;
 #else
 #error "Unsupported host arch"
 #endif
EOF

# ---------------------------------------------------------------------------
# 4. Build the wheel
# ---------------------------------------------------------------------------
if ! find packages/pyvex -name 'pyvex-*-pyodide_*.whl' -type f | grep -q .; then
    echo "[repro] building pyvex wasm32 wheel (~5 min) ..."
    pyodide build-recipes-no-deps pyvex
fi
WHEEL=$(find "$PWD/packages/pyvex" -name 'pyvex-*-pyodide_*.whl' -type f | head -1)
[ -z "$WHEEL" ] && { echo "[repro] BUILD FAILED"; exit 1; }
echo "[repro] wheel built: $WHEEL"

# ---------------------------------------------------------------------------
# 5. The repro Python (identical for native + wasm)
# ---------------------------------------------------------------------------
cat > pyvex-poc.py <<'EOF'
import pyvex, archinfo

# lea  rdi, [rip - 0x14fc7]      48 8d 3d 39 b0 fe ff
# mov  ebp, 0x14                 bd 14 00 00 00
# lea  rdx, [rsp + 0x20]         48 8d 54 24 20
# mov  ecx, dword ptr [rdi - 4]  8b 4f fc            <-- load target = 0x100006320
# call 0x10002e370               e8 73 30 01 00
BYTES = bytes.fromhex(
    "488d3d39b0feff" "bd14000000" "488d542420" "8b4ffc" "e873300100"
)
ADDR = 0x10001b2e4
EXPECTED = 0x100006320

irsb = pyvex.lift(
    BYTES, ADDR, archinfo.ArchAMD64(),
    max_inst=5, collect_data_refs=True,
    opt_level=1, cross_insn_opt=False,
)
print(f"size={irsb.size} jk={irsb.jumpkind} insns={irsb.instructions}")
print("data_refs:")
for r in irsb.data_refs or []:
    print(
        f"  ins=0x{r.ins_addr:x}  data_addr=0x{r.data_addr:x}"
        f"  size={r.data_size}  type={r.data_type_str}"
    )

load_refs = [r for r in (irsb.data_refs or [])
             if r.ins_addr == 0x10001b2f5 and r.data_size == 4]
print()
if not load_refs:
    print("FAIL: no size=4 ref from `mov ecx, [rdi - 4]`")
elif load_refs[0].data_addr == EXPECTED:
    print(f"PASS: data_addr=0x{EXPECTED:x}")
else:
    print(f"FAIL: data_addr=0x{load_refs[0].data_addr:x}, expected 0x{EXPECTED:x}")
EOF

# ---------------------------------------------------------------------------
# 6. Node + Pyodide harness
# ---------------------------------------------------------------------------
mkdir -p harness
if [ ! -d harness/node_modules/pyodide ]; then
    echo "[repro] installing pyodide npm ..."
    ( cd harness && npm init -y >/dev/null && npm install --quiet pyodide@0.29.4 )
fi

cat > harness/run.mjs <<EOF
import {loadPyodide} from 'pyodide';
import {readFileSync} from 'node:fs';

const WHEEL = '$WHEEL';
const POC   = '$WORKDIR/pyvex-poc.py';

(async () => {
    const py = await loadPyodide({stdout: l => process.stdout.write(l + '\\n')});
    await py.loadPackage(['micropip']);
    await py.runPythonAsync(\`
import micropip
await micropip.install([
    "file://\${WHEEL}",
    "archinfo==$ARCHINFO_VERSION",
])
\`);
    await py.runPythonAsync(readFileSync(POC, 'utf-8'));
})();
EOF

# ---------------------------------------------------------------------------
# 7. Run native + wasm and compare
# ---------------------------------------------------------------------------
if [ ! -d native-venv ]; then
    python3.13 -m venv native-venv
    native-venv/bin/pip install --quiet \
        "pyvex==$PYVEX_VERSION" "archinfo==$ARCHINFO_VERSION"
fi

echo
echo "============================================================"
echo "  NATIVE (Linux x86-64, pyvex $PYVEX_VERSION from PyPI)"
echo "============================================================"
native-venv/bin/python pyvex-poc.py

echo
echo "============================================================"
echo "  WASM   (wasm32-emscripten, pyvex $PYVEX_VERSION from local wheel)"
echo "============================================================"
node harness/run.mjs
```

### Environment

_No response_

### Additional context

_No response_

demoray:
Fixes #539.

## Summary

`pyvex_c/analysis.c`'s `HashHW` table — used to cache (IR-temp /
register → resolved-constant) bindings while walking an IRSB to
populate `IRSB.data_refs` — stores values as `HWord` (= `unsigned
long`, "same size as `void*` on the host", per
`vex/pub/libvex_basictypes.h:160-168`). On 64-bit `x86_64-linux-gnu`
this is 64 bits and the bug is invisible because the host width happens
to equal the AMD64 guest width. On `wasm32-emscripten`, `unsigned long`
is 32 bits but the guest stays 64-bit, so the high 32 bits of any
cached guest pointer are silently dropped. The next `lookupHHW` returns
the truncated value and `data_refs` reports the wrong `data_addr`.

The keys are pyvex-internal identifiers (IR-temp numbers, packed
register-offset/type), not pointers, so leaving them at `HWord` is
fine — only the value width matters.

## Reproduction

The downstream symptom shows up on any 64-bit PE image with
`ImageBase >= 0x1_00000000` analyzed under
`angr.analyses.CFGFast(data_references=True)` on a 32-bit host build of
pyvex. Self-contained 25-byte AMD64 PoC (see also #539):

```python
import pyvex, archinfo

# lea  rdi, [rip - 0x14fc7]      48 8d 3d 39 b0 fe ff
# mov  ebp, 0x14                 bd 14 00 00 00
# lea  rdx, [rsp + 0x20]         48 8d 54 24 20
# mov  ecx, dword ptr [rdi - 4]  8b 4f fc            <- target = 0x100006320
# call 0x10002e370               e8 73 30 01 00
BYTES = bytes.fromhex("488d3d39b0feffbd14000000488d5424208b4ffce873300100")
ADDR = 0x10001b2e4

irsb = pyvex.lift(BYTES, ADDR, archinfo.ArchAMD64(),
                  max_inst=5, collect_data_refs=True,
                  opt_level=1, cross_insn_opt=False)

[r for r in irsb.data_refs
   if r.ins_addr == 0x10001b2f5 and r.data_size == 4][0].data_addr
# x86_64-linux-gnu: 0x100006320   (correct)
# wasm32-emscripten: 0x6320       (low 32 bits of 0x100006320)
```

The previous-instruction LEA resolves `rdi = 0x100006324` correctly in
both builds (that data_ref is fine). It's the value flowing from
`tmps[].value` through `addToHHW` into the next instruction's load
resolution that gets clipped.

## Fix

Six small changes in `pyvex_c/analysis.c`:

- `HashHW.val` typed `ULong*` instead of `HWord*`.
- `newHHW` / array-resize allocations use `sizeof(ULong)`.
- `lookupHHW` out-param is `ULong*`.
- `addToHHW` value parameter is `ULong`.
- Local `HWord val` in the `Iex_Get` arm of `process_statement_for_data_refs`
  becomes `ULong val`.

The callers (`addToHHW(env, key, tmps[...].value)`,
`addToHHW(env, key, get_value_from_const_expr(...))`, etc.) already
pass `ULong` — they were being implicitly narrowed at the call site.

## Test

`tests/test_data_refs_64bit_target.py` (new) lifts the snippet above
and asserts `data_addr == 0x100006320` for the size=4 integer ref.
Passes on x86_64-linux-gnu both before and after this PR (the test
documents the expected value); the actual regression catch happens on
a 32-bit host build of pyvex. Verified manually under Pyodide 0.29.4 /
emscripten 4.0.9: `data_addr=0x6320` before, `0x100006320` after.

All 58 existing tests under `tests/` still pass.

## Compatibility

- ABI: `HashHW` is a `static` symbol in `analysis.c`; no callers
  outside this translation unit.
- Memory: 4 extra bytes per HashHW slot on 32-bit hosts. HashHW lives
  for the duration of one IRSB walk and grows powers-of-two from 8
  slots, so the overhead is negligible.
- Behavior on 64-bit hosts: unchanged (`sizeof(ULong) == sizeof(HWord)`).

### DIFF
--- a/pyvex_c/analysis.c
+++ b/pyvex_c/analysis.c
@@ -210,7 +210,7 @@ typedef
    struct {
       Bool*  inuse;
       HWord* key;
-      HWord* val;
+      ULong* val;
       Int    size;
       Int    used;
    }
@@ -223,7 +223,7 @@ static HashHW* newHHW()
    h->used   = 0;
    h->inuse  = (Bool*)malloc(h->size * sizeof(Bool));
    h->key    = (HWord*)malloc(h->size * sizeof(HWord));
-   h->val    = (HWord*)malloc(h->size * sizeof(HWord));
+   h->val    = (ULong*)malloc(h->size * sizeof(ULong));
    return h;
 }
 
@@ -238,7 +238,7 @@ static void freeHHW(HashHW* h)
 
 /* Look up key in the map. */
 
-static Bool lookupHHW(HashHW* h, /*OUT*/HWord* val, HWord key)
+static Bool lookupHHW(HashHW* h, /*OUT*/ULong* val, HWord key)
 {
    Int i;
 
@@ -255,7 +255,7 @@ static Bool lookupHHW(HashHW* h, /*OUT*/HWord* val, HWord key)
 
 /* Add key->val to the map.  Replaces any existing binding for key. */
 
-static void addToHHW(HashHW* h, HWord key, HWord val)
+static void addToHHW(HashHW* h, HWord key, ULong val)
 {
    Int i, j;
 
@@ -272,7 +272,7 @@ static void addToHHW(HashHW* h, HWord key, HWord val)
       /* Copy into arrays twice the size. */
       Bool*  inuse2 = malloc(2 * h->size * sizeof(Bool));
       HWord* key2   = malloc(2 * h->size * sizeof(HWord));
-      HWord* val2   = malloc(2 * h->size * sizeof(HWord));
+      ULong* val2   = malloc(2 * h->size * sizeof(ULong));
       for (i = j = 0; i < h->size; i++) {
          if (!h->inuse[i]) continue;
          inuse2[j] = True;
@@ -811,7 +811,7 @@ void execute_irsb(
 				case Iex_Get:
 					{
 						UInt key = mk_key_GetPut(data->Iex.Get.offset, data->Iex.Get.ty);
-						HWord val;
+						ULong val;
 						if (lookupHHW(env, &val, key) == True) {
 							tmps[stmt->Ist.WrTmp.tmp].used = 1;
 							tmps[stmt->Ist.WrTmp.tmp].value = val;
--- a/tests/test_data_refs_64bit_target.py
+++ b/tests/test_data_refs_64bit_target.py
@@ -1,0 +1,52 @@
+"""Regression test for HashHW value-slot truncation on 32-bit hosts.
+
+See pyvex_c/analysis.c::HashHW. Before the fix, the cached register
+value flowed through a `HWord` (= `unsigned long`, 32-bit on
+wasm32-emscripten and other ILP32 hosts), silently dropping the high
+32 bits of any guest pointer. The PoC below uses an AMD64 guest LEA +
+memory load to land a 64-bit pointer in the HashHW.
+
+https://github.com/angr/pyvex/issues/539
+"""
+import unittest
+
+import archinfo
+import pyvex
+
+
+class TestDataRefs64BitTarget(unittest.TestCase):
+    def test_64bit_guest_pointer_survives_hashhw_roundtrip(self):
+        # 5 AMD64 instructions, 25 bytes. The mov at +0xe loads from
+        # rdi - 4 where rdi was set by the previous lea to 0x100006324,
+        # so the load target should be 0x100006320.
+        blob = bytes.fromhex(
+            "488d3d39b0feff"        # lea rdi, [rip - 0x14fc7]
+            "bd14000000"            # mov ebp, 0x14
+            "488d542420"            # lea rdx, [rsp + 0x20]
+            "8b4ffc"                # mov ecx, dword ptr [rdi - 4]
+            "e873300100"            # call rel32
+        )
+        irsb = pyvex.lift(
+            blob,
+            0x10001B2E4,
+            archinfo.ArchAMD64(),
+            max_inst=5,
+            collect_data_refs=True,
+            opt_level=1,
+            cross_insn_opt=False,
+        )
+        sized_refs = [
+            r for r in (irsb.data_refs or [])
+            if r.ins_addr == 0x10001B2F5 and r.data_size == 4
+        ]
+        assert sized_refs, "expected one size=4 ref from `mov ecx, [rdi - 4]`"
+        self.assertEqual(
+            sized_refs[0].data_addr,
+            0x100006320,
+            f"data_addr=0x{sized_refs[0].data_addr:x}; "
+            "high 32 bits dropped — see pyvex_c/analysis.c HashHW value width",
+        )
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_data_refs_64bit_target.py
+++ b/tests/test_data_refs_64bit_target.py
@@ -11,12 +11,12 @@ https://github.com/angr/pyvex/issues/539
 
 import unittest
 
-import archinfo
-
 import pyvex
 
 
 class TestDataRefs64BitTarget(unittest.TestCase):
+    """Regression tests for HashHW value-slot width on 32-bit hosts."""
+
     def test_64bit_guest_pointer_survives_hashhw_roundtrip(self):
         # 5 AMD64 instructions, 25 bytes. The mov at +0xe loads from
         # rdi - 4 where rdi was set by the previous lea to 0x100006324,
@@ -31,7 +31,7 @@ class TestDataRefs64BitTarget(unittest.TestCase):
         irsb = pyvex.lift(
             blob,
             0x10001B2E4,
-            archinfo.ArchAMD64(),
+            pyvex.ARCH_AMD64,
             max_inst=5,
             collect_data_refs=True,
             opt_level=1,
--- a/tests/test_data_refs_64bit_target.py
+++ b/tests/test_data_refs_64bit_target.py
@@ -8,9 +8,11 @@ memory load to land a 64-bit pointer in the HashHW.
 
 https://github.com/angr/pyvex/issues/539
 """
+
 import unittest
 
 import archinfo
+
 import pyvex
 
 
@@ -20,11 +22,11 @@ class TestDataRefs64BitTarget(unittest.TestCase):
         # rdi - 4 where rdi was set by the previous lea to 0x100006324,
         # so the load target should be 0x100006320.
         blob = bytes.fromhex(
-            "488d3d39b0feff"        # lea rdi, [rip - 0x14fc7]
-            "bd14000000"            # mov ebp, 0x14
-            "488d542420"            # lea rdx, [rsp + 0x20]
-            "8b4ffc"                # mov ecx, dword ptr [rdi - 4]
-            "e873300100"            # call rel32
+            "488d3d39b0feff"  # lea rdi, [rip - 0x14fc7]
+            "bd14000000"  # mov ebp, 0x14
+            "488d542420"  # lea rdx, [rsp + 0x20]
+            "8b4ffc"  # mov ecx, dword ptr [rdi - 4]
+            "e873300100"  # call rel32
         )
         irsb = pyvex.lift(
             blob,
@@ -35,10 +37,7 @@ class TestDataRefs64BitTarget(unittest.TestCase):
             opt_level=1,
             cross_insn_opt=False,
         )
-        sized_refs = [
-            r for r in (irsb.data_refs or [])
-            if r.ins_addr == 0x10001B2F5 and r.data_size == 4
-        ]
+        sized_refs = [r for r in (irsb.data_refs or []) if r.ins_addr == 0x10001B2F5 and r.data_size == 4]
         assert sized_refs, "expected one size=4 ref from `mov ecx, [rdi - 4]`"
         self.assertEqual(
             sized_refs[0].data_addr,


rhelmot:
lgtm, I'll merge when CI passes

demoray:
> lgtm, I'll merge when CI passes

Thanks.  I believe I've fixed the currently shown issues from CI and will check back tomorrow to verify nothing else cropped up.

PR #557, opened 2026-07-22
Analysis: Support 64-bit data loading; Fix region ordering corruption.

ltfish:


ltfish:
The failed test case is addressed in https://github.com/angr/angr/pull/6667.

PR #564, opened 2026-08-11
Lifting: Tell libVEX how much of the lift buffer it may read.

### LINKED ISSUE #561: Segmentation fault lifting a truncated s390x instruction

## Summary

Lifting seven bytes of s390x code segfaults the interpreter. There is no Python
exception — the process dies with SIGSEGV inside libVEX:

```text
Fatal Python error: Segmentation fault

Current thread (most recent call first):
  File "pyvex/lifting/libvex.py", line 87 in _lift
  File "pyvex/lifting/lifter.py", line 110 in lift
  File "pyvex/lifting/lift_function.py", line 134 in lift
```

`lift_function.py` already pads byte input with eight zero bytes specifically to
give libVEX room to over-read, so the intent to tolerate a truncated final
instruction is there. The s390x front end reads past even that padding when the
block ends mid-instruction: the input below is a complete `a75b ffff` followed by
`c6 50 ff`, the first three bytes of a six-byte RIL-format instruction.

Because it is a segfault rather than an exception, a caller cannot defend against
it — the whole process is lost, which matters for anything lifting untrusted or
partially-mapped code.

## Reproduction

Deterministic, and independent of the load address:

```python
import pyvex
import archinfo

pyvex.lift(bytes.fromhex("a75bffffc650ff"), 0x1000, archinfo.arch_from_id("s390x"))
```

```console
$ python repro.py; echo "exit=$?"
Fatal Python error: Segmentation fault
exit=139
```

Tested with pyvex `26f2b61c618bdd4e01be44562c97b824be399200` and archinfo
`dad25b035d090c9165e85a65f16322456568df43`. Reached in practice through
`CFGFast`, which lifts at addresses it discovers by scanning and so routinely
hands the lifter a block that ends mid-instruction.

## Scope

Found by a corpus sweep running `CFGFast` over every unit that loads. Native
crashes account for 81 VEX-lifted units at 30% of that sweep, all real-world
material and concentrated in stock Debian port binaries; s390x is the largest
group among them.

The audit records these as a worker killed by a signal, with no Python
traceback, so the true total may be larger — a native crash cannot be attributed
to a call site from inside the dead process.

zardus:

### Problem

Lifting s390x takes the interpreter down, with no exception to catch. Six
well-formed bytes are enough — `c660ffffffff` is `exrl %r6,.-2`, the backward
form compilers emit for literal-pool-relative `MVC` and `CLC`:

```
$ pyvex.lift(bytes.fromhex("c660ffffffff"), 0x1000, archinfo.ArchS390X())
  [killed by SIGSEGV]
$ pyvex.lift(bytes.fromhex("c66000000003"), 0x1000, archinfo.ArchS390X())
  NEXT: PUT(ia) = 0x0000000000001000; Ijk_InvalICache
```

The sign of the displacement is the whole difference. `a75bffffc650ff` — an
`aghi` followed by a truncated `exrl`, which is what a scanning CFG produces —
dies the same way.

### Root cause

libVEX's s390x front end reads the `EXRL` execute target out of the buffer it
was handed, at a displacement nothing bounds; the vex branch this bumps to
bounds that read.

The half this repository owns is that pyvex never told libVEX how large that
buffer is. `pvc.vex_lift` passes `max_bytes`, which is the size of the block to
translate, not the size of the readable region: angr hands over a whole segment
and caps the block at `VEX_IRSB_MAX_SIZE`, 400 bytes, and real `EXRL` targets
sit past that cap, so bounding the prefetch by `max_bytes` alone would drop
legitimate targets. pyvex did know the extent and threw it away:
`lift_function.py` cast the cffi buffer through `ffi.BVoidP`, and a `void *`
carries no size.

```
>>> ffi.sizeof(ffi.from_buffer(ffi.BVoidP, bytearray(64)))   # <ctype 'void *'>
8
>>> ffi.sizeof(ffi.from_buffer(bytearray(64)))               # <ctype 'char[]'>
64
```

### Fix

Keep the sized `char[]` and state its extent. `libvex.py` computes
`ffi.sizeof(self.data) - self.bytes_offset` when the object is an array, zero
when it is a bare pointer, and passes it to a new `vex_lift_with_buffer_size`,
which sets `vc.guest_bytes_size`. `vex_lift` keeps its signature and forwards
zero — unstated — so angr's native consumers build and run unchanged. The
submodule bump points at the vex branch that consumes the field.

### Testing

`tests/test_s390x_exrl.py` covers a target before the buffer, one past it, one
past `max_bytes` but still inside the buffer, one reached through a non-zero
`bytes_offset`, and a truncated encoding. All six pass on this head, five runs
of five. Without the fix two are killed by SIGSEGV and two fail their
assertions. One of those two, `test_s390x_exrl_target_past_buffer`, reads
uninitialised heap 8 KiB past the buffer, so what it finds is not guaranteed:
unfixed it failed 236 of 240 fresh interpreters and passed four.

A seventh test lifts real compiler output instead of a byte string:
`tests/s390x/exrl/acme-client`, the Alpine v3.23 s390x build of acme-client
1.3.5 that went into angr/binaries as pull request 222, at `0x40723a`. That is where `CFGFast`'s scan of the
executable segment lands -- two bytes into the word the entry stub loads to find
`_DYNAMIC` -- and the six bytes there decode as `exrl %r6` aiming 235,802,126
bytes ahead, about 225 MiB past the end of the 73,920-byte segment cle hands
over. It is killed by SIGSEGV in forty fresh interpreters out of forty without
the fix, and `CFGFast` over the whole file is killed on pyvex master three
runs out of three, where this head recovers 435 functions. Sites already in
`tests/s390x` crash the same way when lifted at a fixed offset -- `ld64.so.1`
at `0x1304`, say -- but a scan starts a block at none of the seventy such sites
in the repository, so the new object is the only one where the address a test
pins is one an analysis really reaches. angr/binaries is a sibling only in the
`ecosystem` job; the standalone platform jobs check out pyvex alone, and the
test skips there naming the directory it wanted.

Needs angr/vex#89, which the submodule bump here points at.

Fixes #561. Validation: https://github.com/angr/pyvex/pull/564#issuecomment-5257392181

Merge order: angr/vex#89 goes in before this one. This head's `vex` submodule is `88aa12d`, which is that pull request's head and one commit ahead of `vex` master, so merging this half on its own leaves pyvex master's submodule pointing at a commit that is on no vex branch. The fixture half has already landed -- binaries master carries `tests/s390x/exrl/acme-client` since 2026-08-31 -- so the seventh test loads it from there and no reference to that repository is needed here. `angr/vex` is not in the CI repository list, so the `sync:` line below is a note to the reader rather than an instruction to CI.

sync: angr/vex#89

session: sharpen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,8 @@ add_custom_command(
     COMMAND ${CMAKE_COMMAND} -E env
         ${Python3_EXECUTABLE} ${CMAKE_SOURCE_DIR}/make_ffi.py ${CMAKE_SOURCE_DIR}/vex/pub
     DEPENDS ${CMAKE_SOURCE_DIR}/vex/pub/libvex_guest_offsets.h
+            ${CMAKE_SOURCE_DIR}/vex/pub/libvex.h
+            ${CMAKE_SOURCE_DIR}/pyvex_c/pyvex.h
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     COMMENT "Generating pyvex/vex_ffi.py using make_ffi.py"
 )
--- a/pyvex/lifting/libvex.py
+++ b/pyvex/lifting/libvex.py
@@ -31,6 +31,8 @@ LIBVEX_SUPPORTED_ARCHES = {
 
 VEX_MAX_INSTRUCTIONS = 99
 VEX_MAX_BYTES = 5000
+# libVEX stores the buffer size in an Int.
+VEX_MAX_BUFFER_SIZE = 0x7FFFFFFF
 
 
 class VexRegisterUpdates:
@@ -80,19 +82,29 @@ class LibVEXLifter(Lifter):
             if strict_block_end is None:
                 strict_block_end = True
 
+            # The s390x EXRL target is read straight out of the buffer, and it
+            # can sit well past max_bytes, so libVEX needs to know how much of
+            # the buffer is there. A bare pointer carries no size; leave it
+            # unstated then, and libVEX reads no further than max_bytes.
+            if ffi.typeof(self.data).kind == "array":
+                buffer_size = min(max(ffi.sizeof(self.data) - self.bytes_offset, 0), VEX_MAX_BUFFER_SIZE)
+            else:
+                buffer_size = 0
+
             if self.cross_insn_opt:
                 px_control = VexRegisterUpdates.VexRegUpdUnwindregsAtMemAccess
             else:
                 px_control = VexRegisterUpdates.VexRegUpdLdAllregsAtEachInsn
 
             vex_archinfo["hwcache_info"]["caches"] = ffi.NULL
-            lift_r = pvc.vex_lift(
+            lift_r = pvc.vex_lift_with_buffer_size(
                 vex_arch,
                 vex_archinfo,
                 self.data + self.bytes_offset,
                 self.irsb.addr,
                 max_inst,
                 max_bytes,
+                buffer_size,
                 self.opt_level,
                 self.traceflags,
                 self.allow_arch_optimizations,
--- a/pyvex/lifting/lift_function.py
+++ b/pyvex/lifting/lift_function.py
@@ -112,10 +112,12 @@ def lift(
                     # near) it. Lift from a copy padded with 8 NUL bytes so
                     # such reads are deterministic instead of heap garbage;
                     # keep the zero-copy path when there is enough slack.
+                    # The result is a sized char[]; libvex.py passes that size
+                    # on for decoders that read outside the block.
                     if isinstance(py_data, (bytearray, memoryview)) and bytes_offset + max_bytes + 8 <= len(py_data):
-                        u_data = ffi.from_buffer(ffi.BVoidP, py_data)
+                        u_data = ffi.from_buffer(py_data)
                     else:
-                        u_data = ffi.from_buffer(ffi.BVoidP, bytes(py_data) + b"\0" * 8)
+                        u_data = ffi.from_buffer(bytes(py_data) + b"\0" * 8)
                 else:
                     u_data = c_data
                 skip = 0
--- a/pyvex_c/pyvex.c
+++ b/pyvex_c/pyvex.c
@@ -329,6 +329,28 @@ VEXLiftResult *vex_lift(
 		int const_prop,
 		VexRegisterUpdates px_control,
 		unsigned int lookback) {
+	return vex_lift_with_buffer_size(guest, archinfo, insn_start, insn_addr, max_insns, max_bytes,
+			0 /* buffer_size */, opt_level, traceflags, allow_arch_optimizations, strict_block_end,
+			collect_data_refs, load_from_ro_regions, const_prop, px_control, lookback);
+}
+
+VEXLiftResult *vex_lift_with_buffer_size(
+		VexArch guest,
+		VexArchInfo archinfo,
+		unsigned char *insn_start,
+		unsigned long long insn_addr,
+		unsigned int max_insns,
+		unsigned int max_bytes,
+		unsigned int buffer_size,
+		int opt_level,
+		int traceflags,
+		int allow_arch_optimizations,
+		int strict_block_end,
+		int collect_data_refs,
+		int load_from_ro_regions,
+		int const_prop,
+		VexRegisterUpdates px_control,
+		unsigned int lookback) {
 	VexRegisterUpdates pxControl = px_control;
 
 	vex_prepare_vai(guest, &archinfo);
@@ -346,6 +368,7 @@ VEXLiftResult *vex_lift(
 	vta.traceflags          = traceflags;
 
 	vc.guest_max_bytes     = max_bytes;
+	vc.guest_bytes_size    = buffer_size;
 	vc.guest_max_insns     = max_insns;
 	vc.iropt_level         = opt_level;
 	vc.lookback_amount     = lookback;
--- a/pyvex_c/pyvex.def
+++ b/pyvex_c/pyvex.def
@@ -48,6 +48,7 @@ EXPORTS
   typeOfPrimop
   clear_log
   vex_lift
+  vex_lift_with_buffer_size
   vex_init
   register_readonly_region
   deregister_all_readonly_regions
--- a/pyvex_c/pyvex.h
+++ b/pyvex_c/pyvex.h
@@ -88,6 +88,28 @@ VEXLiftResult *vex_lift(
 		VexRegisterUpdates px_control,
 		unsigned int lookback_amount);
 
+// As vex_lift, but the caller also states how many bytes are readable at
+// insn_start. A decoder that looks outside the block it is lifting -- the
+// s390x EXRL target is read out of the buffer -- stays inside that. Zero
+// means unstated, in which case no more than max_bytes is read.
+VEXLiftResult *vex_lift_with_buffer_size(
+		VexArch guest,
+		VexArchInfo archinfo,
+		unsigned char *insn_start,
+		unsigned long long insn_addr,
+		unsigned int max_insns,
+		unsigned int max_bytes,
+		unsigned int buffer_size,
+		int opt_level,
+		int traceflags,
+		int allow_arch_optimizations,
+		int strict_block_end,
+		int collect_data_refs,
+		int load_from_ro_regions,
+		int const_prop,
+		VexRegisterUpdates px_control,
+		unsigned int lookback_amount);
+
 Bool register_readonly_region(ULong start, ULong size, unsigned char* content);
 void deregister_all_readonly_regions();
 Bool register_initial_register_value(UInt offset, UInt size, ULong value);
--- a/tests/test_s390x_exrl.py
+++ b/tests/test_s390x_exrl.py
@@ -21,5 +21,119 @@ def test_s390x_exrl():
     assert irsb.jumpkind == "Ijk_Ret"
 
 
+def test_s390x_exrl_target_before_buffer():
+    """The EXRL displacement is signed. Reading it as unsigned aimed the target
+    about 8GB past the buffer, and put that address in the emitted lookup too.
+    """
+    arch = pyvex.ARCH_S390X
+    irsb = pyvex.lift(
+        b"\xc6\x10\xff\xff\xff\x00"  # exrl %r1,0x400200
+        b"\x07\xfe",  # br %r14
+        0x400400,
+        arch,
+    )
+    irsb_str = str(irsb)
+
+    # The target is 0x200 bytes back, so it is not in the buffer and cannot be
+    # prefetched; the run-time lookup is emitted, and it reads the right place.
+    assert "LDbe:I64(0x0000000000400200)" in irsb_str
+    assert "s390x_dirtyhelper_EX" in irsb_str
+    assert irsb.size == 6
+    assert irsb.jumpkind == "Ijk_InvalICache"
+
+
+def test_s390x_exrl_target_past_buffer():
+    """Nothing constrains the displacement, so a bogus EXRL points anywhere
+    within 4GB of the buffer. Nothing outside it is read; the run-time lookup
+    resolves the target instead.
+    """
+    arch = pyvex.ARCH_S390X
+    irsb = pyvex.lift(
+        b"\xc6\x10\x00\x00\x10\x00"  # exrl %r1,0x402400
+        b"\x07\xfe",  # br %r14
+        0x400400,
+        arch,
+    )
+    irsb_str = str(irsb)
+
+    assert "LDbe:I64(0x0000000000402400)" in irsb_str
+    assert "s390x_dirtyhelper_EX" in irsb_str
+    assert irsb.size == 6
+    assert irsb.jumpkind == "Ijk_InvalICache"
+
+
+def test_s390x_exrl_target_past_max_bytes():
+    """An EXRL target may sit outside the lift window and still be inside the
+    caller's buffer, which is how angr lifts: it hands over a whole segment and
+    caps the block at max_bytes. The target is readable, so it is still used.
+    """
+    arch = pyvex.ARCH_S390X
+    block = (
+        b"\xc6\x10\x00\x00\x00\x64"  # exrl %r1,0x4004c8
+        b"\x07\xfe"  # br %r14
+    )
+    target = b"\xd7\x00\x20\x00\x30\x00"  # xc 0(0,%r2),0(%r3)
+    buf = bytearray(block + b"\x00" * (0xC8 - len(block)) + target)
+
+    for data in (bytes(buf), pyvex.ffi.from_buffer(buf)):
+        irsb = pyvex.lift(data, 0x400400, arch, max_bytes=len(block))
+        assert "0xd700200030000000" in str(irsb)
+        assert irsb.jumpkind == "Ijk_Ret"
+
+    # Drop the last byte of the target and it no longer fits in the buffer.
+    irsb = pyvex.lift(pyvex.ffi.from_buffer(buf[:-1]), 0x400400, arch, max_bytes=len(block))
+    assert "0xd700200030000000" not in str(irsb)
+    assert irsb.jumpkind == "Ijk_InvalICache"
+
+    # A bare pointer has no extent to state, so max_bytes is all that is read.
+    sized = pyvex.ffi.from_buffer(buf)
+    irsb = pyvex.lift(pyvex.ffi.cast("unsigned char *", sized), 0x400400, arch, max_bytes=len(block))
+    assert "0xd700200030000000" not in str(irsb)
+    assert irsb.jumpkind == "Ijk_InvalICache"
+
+
+def test_s390x_exrl_inside_a_larger_buffer():
+    """The whole segment angr hands over starts before the block, and the EXRL
+    need not be the block's first instruction, so the target is displaced from
+    the EXRL rather than from either start.
+    """
+    arch = pyvex.ARCH_S390X
+    block = (
+        b"\xa7\x5b\xff\xff"  # aghi %r5,-1
+        b"\xc6\x10\x00\x00\x00\x62"  # exrl %r1,0x4004c8
+        b"\x07\xfe"  # br %r14
+    )
+    target = b"\xd7\x00\x20\x00\x30\x00"  # xc 0(0,%r2),0(%r3)
+    prefix = b"\x07\xfe" * 8  # code before the block, never decoded
+    buf = bytearray(prefix + block + b"\x00" * (0xC8 - len(block)) + target)
+
+    irsb = pyvex.lift(pyvex.ffi.from_buffer(buf), 0x400400, arch, max_bytes=len(block), bytes_offset=len(prefix))
+    assert "0xd700200030000000" in str(irsb)
+    assert irsb.jumpkind == "Ijk_Ret"
+
+
+def test_s390x_exrl_truncated():
+    """A block whose last instruction is a truncated EXRL. CFGFast lifts at
+    addresses it finds by scanning, so it hands the lifter such a block
+    routinely.
+    """
+    arch = pyvex.ARCH_S390X
+    irsb = pyvex.lift(
+        b"\xa7\x5b\xff\xff"  # aghi %r5,-1
+        b"\xc6\x50\xff",  # first three bytes of a six-byte exrl
+        0x1000,
+        arch,
+    )
+
+    # The exrl does not fit in the block, so only the aghi is lifted.
+    assert irsb.size == 4
+    assert irsb.jumpkind == "Ijk_Boring"
+
+
 if __name__ == "__main__":
     test_s390x_exrl()
+    test_s390x_exrl_target_before_buffer()
+    test_s390x_exrl_target_past_buffer()
+    test_s390x_exrl_target_past_max_bytes()
+    test_s390x_exrl_inside_a_larger_buffer()
+    test_s390x_exrl_truncated()
--- a/tests/test_s390x_exrl.py
+++ b/tests/test_s390x_exrl.py
@@ -1,4 +1,30 @@
+import os
+import unittest
+
 import pyvex
+
+# angr/binaries is checked out beside the repository under test in the
+# ecosystem CI job. The standalone platform jobs check out pyvex alone, so the
+# real-binary test below skips there.
+BINARIES = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "binaries", "tests")
+ACME_CLIENT = os.path.join(BINARIES, "s390x", "exrl", "acme-client")
+
+# angr maps that PIE at 0x400000 and CFGFast scans the executable segment for
+# code, which lands it on 0x40723a. Nothing executes there: the entry stub ends
+# at 0x407233 and branches to 0x407240, and 0x40723a is two bytes into the word
+# at 0x407238 that the stub loads to find _DYNAMIC. Its low half plus the
+# padding after it read as exrl %r6 with a target 235,802,126 bytes ahead.
+# tests/s390x/exrl/README.md in angr/binaries has the rest.
+LOAD_BASE = 0x400000
+ENTRY_OFFSET = 0x7210
+EXRL_OFFSET = 0x723A
+
+
+def read_acme_client():
+    if not os.path.isdir(BINARIES):
+        raise unittest.SkipTest(f"angr/binaries is not checked out beside pyvex: {BINARIES} is missing")
+    with open(ACME_CLIENT, "rb") as fp:
+        return fp.read()
 
 
 def test_s390x_exrl():
@@ -130,6 +156,39 @@ def test_s390x_exrl_truncated():
     assert irsb.jumpkind == "Ijk_Boring"
 
 
+def test_s390x_exrl_real_binary():
+    """A compiler-produced object that this used to kill the process on:
+    acme-client 1.3.5 for s390x, out of Alpine v3.23. CFGFast's scan reaches a
+    data word between two functions, the six bytes there decode as an EXRL
+    aiming 235,802,126 bytes ahead, and libVEX fetched that out of the
+    translation buffer without a bounds check -- around 225 MiB past the end of
+    the 73,920-byte executable segment cle hands over.
+    """
+    image = read_acme_client()
+    arch = pyvex.ARCH_S390X
+
+    # The entry stub, lifted out of the file. Nothing in it reads outside the
+    # buffer, so it says the fixture is being read and lifted correctly. Its
+    # last instruction is the branch to 0x407240, which is why nothing ever
+    # runs the bytes the scan trips on.
+    irsb = pyvex.lift(image, LOAD_BASE + ENTRY_OFFSET, arch, bytes_offset=ENTRY_OFFSET)
+    assert irsb.size == 36
+    assert irsb.jumpkind == "Ijk_Boring"
+    assert "PUT(ia) = 0x0000000000407240; Ijk_Boring" in str(irsb)
+
+    # And the address the scan reaches, both as the whole file and as the six
+    # bytes a block cut at max_bytes leaves. The target is far outside the
+    # buffer either way, so it is not prefetched and the run-time lookup is
+    # emitted instead.
+    for data, offset in ((image, EXRL_OFFSET), (image[EXRL_OFFSET : EXRL_OFFSET + 6], 0)):
+        irsb = pyvex.lift(data, LOAD_BASE + EXRL_OFFSET, arch, bytes_offset=offset)
+        irsb_str = str(irsb)
+        assert "LDbe:I64(0x000000000e4e8048)" in irsb_str
+        assert "s390x_dirtyhelper_EX" in irsb_str
+        assert irsb.size == 6
+        assert irsb.jumpkind == "Ijk_InvalICache"
+
+
 if __name__ == "__main__":
     test_s390x_exrl()
     test_s390x_exrl_target_before_buffer()
@@ -137,3 +196,4 @@ if __name__ == "__main__":
     test_s390x_exrl_target_past_max_bytes()
     test_s390x_exrl_inside_a_larger_buffer()
     test_s390x_exrl_truncated()
+    test_s390x_exrl_real_binary()


zardus:

Validation record for head `ef34018813b736c3750db979e90c993c7f51c360` (vex submodule `88aa12d54fb50e91635e81c0f3c0bffa1b712691`) against baseline `bdd5441035e02920eaa72c1c3cf9a4f0d572104d` (vex `875f7c9a5f6be621b4f000c29c016e15ddf32207`). CPython 3.12.13, Linux x86-64, GCC 15.2.0.

- Regression: `python -c 'import archinfo, pyvex; pyvex.lift(bytes.fromhex("a75bffffc650ff"), 0x1000, archinfo.arch_from_id("s390x"))'` — baseline dies with SIGSEGV, exit 139, in `LibVEXLifter._lift` on the `pvc.vex_lift` call; head returns a 4-byte `Ijk_Boring` block
- Reproducer from ordinary compiler output rather than a truncated instruction: `pyvex.lift(bytes.fromhex("c660ffffffff"), 0x1000, archinfo.ArchS390X())` takes the process down on the baseline and on the released 9.3.3 wheel, while `c66000000003` — the same instruction with a forward displacement — lifts to `Ijk_InvalICache`
- Fixture survey, one process per site, over a narrower site set than the table further down (337 here against 1,309 there; the two scans differ in what they count as a site, not in what they found): over the 337 EXRL sites in `binaries/tests/s390x`, `bdd5441` lifts 161 and is killed on 176, and this branch lifts all 337 and is killed on none. Of the 161 that already lifted, 11 produce different IR here, every difference a correction of a prefetch taken from outside the buffer. Those fixtures include `libc.so.6`, `libstdc++.so.6`, `libm.so.6`, `ld64.so.1`, `libgcc_s.so.1` and `test-instr_s390x`; detail in https://github.com/angr/pyvex/pull/564#issuecomment-5384335826
- Focused: `pytest tests/test_s390x_exrl.py` — 7 passed on head, with `angr/binaries` checked out beside pyvex. Against a baseline-built `libpyvex.so`, `test_s390x_exrl_truncated` and `test_s390x_exrl_target_before_buffer` exit 139, `test_s390x_exrl_target_past_max_bytes` fails its assertion, and the new `test_s390x_exrl_real_binary` is killed by signal 11 — under `pytest --forked`, which is how the ecosystem job runs the suite, that one is reported as a failure rather than taking the run down. The other three pass there, which is what they are for — they guard against bounding the prefetch too tightly.
- Full suite: `pytest tests` — 71 passed, 0 skipped, 0 xfailed
- Real binary: `tests/s390x/exrl/acme-client`, lifted at `0x40723a`, the address `CFGFast`'s scan of the executable segment reaches. Unfixed pyvex `26549469645736c27e81223ba7cb5881e8002e20` (vex `875f7c9`) is killed by SIGSEGV three times out of three, both when handed the whole file with `bytes_offset` and when handed the six bytes alone; this head returns a six-byte `Ijk_InvalICache` block carrying `LDbe:I64(0x000000000e4e8048)`, identical across three runs. The entry stub at `0x407210` lifts to the same 36-byte `Ijk_Boring` block on both revisions, so both arms are reading the same fixture. `CFGFast(normalize=True)` with `auto_load_libs=False` over the whole object: SIGSEGV three of three on that baseline, and 435 functions and 3,669 nodes in 0.9–1.3 s three of three here.
- In the command the ecosystem job runs — `pytest -v -nauto --forked --splits 10 --group <W> --rootdir=./src/pyvex/tests ./src/pyvex/tests -m "not slow"` — the new test is collected and, against a baseline-built `libpyvex.so`, reported `FAILED ... Fatal Python error: Segmentation fault`. Collecting the whole `tests` directory, pytest-split with no `.test_durations` puts the entire file in group 9 with six tests and with seven, so it runs in `ecosystem / Test (8)`; that is where the six ran on the last hosted run. The grep for it still has to cover all ten shards, because nothing pins the split.
- All seven tests at `bdd5441`, the revision this branch was cut from and the one every figure above was measured against: `test_s390x_exrl_real_binary`, `test_s390x_exrl_target_before_buffer` and `test_s390x_exrl_truncated` killed by signal 11, `test_s390x_exrl_target_past_max_bytes` and `test_s390x_exrl_target_past_buffer` `AssertionError`, the other two clean. `test_s390x_exrl_target_past_buffer` reads uninitialised heap, so its outcome is decided by the process it runs in and repeated forks of one parent all agree, which makes a run of forks a single observation. Measured one **fresh interpreter** per run it failed 236 of 240 there and passed four, and passed 20 of 20 at this head; a second 240-run pass in the same session saw no passes at all, so the direction is reproducible and the rate is not. The four passes in five recorded in the before/after comment on 2026-08-28 are a lucky run, not the usual outcome. `test_s390x_exrl_real_binary` in the same harness: signal 11 forty times of forty there, clean twenty of twenty here. At this head all seven are clean five of five.
- Fixture guard: with no `binaries` directory beside pyvex the file reports 6 passed, 1 skipped, and the skip names the path it wanted — that is the standalone `Test` matrix and `Test (Pyodide)`. With a `binaries` checkout that does not carry the fixture it reports 1 failed with `FileNotFoundError`, so a missing cross-repository reference reddens the ecosystem job rather than skipping in silence.
- Lint/format: `pre-commit run --all-files` — 24 hooks, 22 passed, 2 skipped for having no matching files
- Lint/type, scored the way the hosted jobs score them, re-run after the rebase against `26549469645736c27e81223ba7cb5881e8002e20` for both bases, over the 3 changed Python files: pylint 9.18 → 9.23 on `pyvex/lifting/libvex.py` and 10.00 → 10.00 on the other two; pyright badness 0.00 → 0.00 on all three
- Workspace gate: not run. Entering the workspace shell reinstalls the shared editable installs and rewrites native libraries that a live corpus sweep has mapped. What ran instead: pyvex's own suite against a separately built head, the merge-base lint and type comparison above, the full `pre-commit` set, and the workspace `test-inputs` check. The angr, cle, claripy and GUI suites did not run locally; the ecosystem job covers them.
- Mutation check: dropping the `exrl_bytes - guest_code` term from the bounds check, or always leaving the extent unstated, each fails a test in `tests/test_s390x_exrl.py`. Without `test_s390x_exrl_inside_a_larger_buffer` the first mutation passes the whole file.

EXRL survey over the 67 s390x ELFs in angr/binaries `12d015e510e2a5fe1217ef061166b7ed7fda7a04`: every six-byte `c6?0` site in every executable segment, 1309 in total, lifted the way angr lifts one — the whole segment as a cffi buffer, `max_bytes=400`, `bytes_offset` at the site — with one fork per site and IRSB text digests compared.

| Outcome | Baseline | Head |
| --- | ---: | ---: |
| Target prefetched, block continues past the EXRL | 157 | 157 |
| `Ijk_InvalICache`, target left to the run-time lookup | 21 | 90 |
| `Ijk_NoDecode` | 1066 | 1062 |
| Killed by SIGSEGV | 65 | 0 |

- The 157 that prefetch are byte-identical on both revisions, so no resolution is lost. The `exrl` at 0xcf204 in `tests/s390x/libc.so.6` executes the `xc` at 0xcf22a on both, for instance.
- All 69 sites that differ had the baseline reading outside the buffer, and all 69 now stop at `Ijk_InvalICache`. Four of them decoded what followed the buffer rather than faulting: the site at 0x16eeae in `tests/s390x/libc.so.6` displaces to 0x1ac9be, past the segment end at 0x198e28.
- Both revisions reproduce their own results run to run, and the head survey reproduces across two independent clean builds.

Cross-architecture check: 27,918 lifts over X86, AMD64, ARMEL, AARCH64, MIPS32, MIPS64, PPC32, PPC64 and S390X, sampling `tests/<arch>/fauxware` and `tests/<arch>/libc.so.6` from the same angr/binaries revision — byte-identical IRSBs on baseline and head.

Caveats: no s390x hardware was involved; this is lifting evidence only. An EXRL whose target lies inside the caller's buffer but before `insn_start` is not prefetched, which the existing `lookback_amount` could allow later; no such site exists in the corpus above, where all 157 real targets are forward.


---

Corpus frequency, from a sweep that has since completed: 746,966 objects analysed, of which 479 killed the worker with a signal rather than an exception. **This defect is 223 of those 479 — 46.6% of every native crash in the run, and the single largest cause.** It is 223 of 7,255 s390x units, so **3.07% of all s390x objects die on it**, against a corpus-wide crash rate of 0.064%.

220 of the 223 are real-world ELF objects rather than synthetic inputs. For comparison, the next largest crash cause is a SLEIGH parse-tree overrun with 115 records, every one of them a fuzz blob.

The six bytes that reproduce it need no corpus and no angr:

```python
pyvex.lift(bytes.fromhex("c6d0ffffff2d"), 0x42dac2, archinfo.ArchS390X())
```

`c6d0ffffff2d` is `exrl %r13, .-422` — ordinary compiler output, which is why the rate is percentage-scale rather than a handful of malformed files. Sweeping the RI2 field one value per process shows the two regimes: displacements up to about 2 MB forward read past the end of a six-byte buffer and silently fabricate an execute target from whatever is there, while anything at or beyond 4 GB faults.

Confirmed still reproducing on the current pin.

**Re-keyed 2026-08-28.** The figures above were measured at `bc8b2245ff1ef9ffa5250b06972e14bd134bcf43` on baseline `90e9094c6217e4e530bf653c06c386d8a6746f82`, which is the head the opening line named until now; the branch is at `fee0c3aa10108fad6070740705123bef370c052b` on `bdd5441035e02920eaa72c1c3cf9a4f0d572104d`. `git range-diff 90e9094c6217e4e530bf653c06c386d8a6746f82..bc8b2245ff1ef9ffa5250b06972e14bd134bcf43 bdd5441035e02920eaa72c1c3cf9a4f0d572104d..fee0c3aa10108fad6070740705123bef370c052b` marks the commit `!` rather than `=` and `git diff bc8b2245ff1ef9ffa5250b06972e14bd134bcf43 fee0c3aa10108fad6070740705123bef370c052b` differs only by master's own advance (5 files changed, 20 insertions(+), 6 deletions(-)). The difference is entirely context lines: master hoisted `self.irsb.arch.vex_archinfo` into a local `vex_archinfo` in `pyvex/lifting/libvex.py` and threaded an optional `vex_archinfo=None` keyword through `lift_function.py` and `lifter.py`. None of this commit's own added or removed lines differ. Every figure above was measured with that keyword unset, where the local resolves to `arch.vex_archinfo` and the call is the one that was measured. Not re-run at the new baseline: the regression needs a native `libpyvex.so` built against the branch's `vex` pin, so the s390x figures above remain the ones measured at `bc8b224` on baseline `90e9094c`.



**CI.** Run [33359113296](https://github.com/angr/pyvex/actions/runs/33359113296) at `293c69a485b50053a6d308f97b92c3fc7ccd3b0f`, the head before the rebase, is the one that matters for the new test, and it shows it ran rather than skipped. `ecosystem / Build` fetched `refs/pull/222/head` for `angr/binaries`, so `src/binaries` carried the fixture, and `ecosystem / Test (8)` printed

```
[gw0] [ 85%] PASSED src/pyvex/tests/test_s390x_exrl.py::test_s390x_exrl_real_binary
```

No other shard mentions that test and none prints `SKIPPED` for it. All five jobs that check out pyvex alone skip it. `Test (ubuntu-24.04)` reports `70 passed, 1 skipped`; `Test (macos-15)`, `Test (macos-15-intel)`, `Test (windows-2022)` and `Test (Pyodide)` report `69 passed, 2 skipped`, the extra one being the pre-existing `test_pyvex.py::test_memory`, which is `skipUnless(sys.platform == "linux")`. The skip this change adds is the one they have in common, and it names the directory it wanted.

The one red check on that run was `ecosystem / Test (0)`, on two **angr** tests, `test_cfg_mips_n32.py::test_n32_prologue_decodes` and `::test_n32_cfg_covers_symbol_functions`, failing with `block.instructions == 0` on `n32_be_static`. The ecosystem job builds pyvex at this branch's own head, and MIPSN32 lifter support landed on pyvex master in #576 after this branch was cut, so the branch handed angr a lifter that cannot decode MIPSN32. Rebasing onto `2654946` is what clears it; nothing in this diff is involved.

Earlier, run [33021248296](https://github.com/angr/pyvex/actions/runs/33021248296) had `ecosystem / Build` die in `_get_pr` in `resolve_refs.py` with `HTTP Error 403: rate limit exceeded`, on the call that fetches this pull request's own metadata — before the body is parsed and before any clone. `_get_pr` sends an `Authorization` header only when `GITHUB_TOKEN` is set, and at the time the shared `build` job passed none. angr/ci-settings#126 fixed that and merged on 2026-08-27; the build step now carries `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`. That run was re-run afterwards and passed.



**Rebased and re-keyed 2026-08-31.** Two moves since the figures above were taken. The head first went from `fee0c3aa10108fad6070740705123bef370c052b` to `293c69a485b50053a6d308f97b92c3fc7ccd3b0f`, adding `test_s390x_exrl_real_binary` and nothing else — 60 lines in `tests/test_s390x_exrl.py`, no production file touched. It was then rebased from `bdd5441035e02920eaa72c1c3cf9a4f0d572104d` onto master `26549469645736c27e81223ba7cb5881e8002e20` to reach `ef34018813b736c3750db979e90c993c7f51c360`; `git range-diff` marks both commits `=`, so the replay changed nothing, and the `vex` pin is still `88aa12d54fb50e91635e81c0f3c0bffa1b712691`. The rebase was needed rather than cosmetic: the ecosystem job builds pyvex at this branch's own head, and master had gained MIPSN32 lifter support (#576) that angr master's `test_cfg_mips_n32` needs, so `ecosystem / Test (0)` failed two angr tests at `293c69a` with `block.instructions == 0` on `n32_be_static`. Those two commits touch `pyproject.toml`, `pyvex/__init__.py`, `pyvex/arches.py`, one line of `pyvex/lifting/libvex.py` and a new `tests/test_mipsn32.py`; one of them is `pyvex/lifting/libvex.py`, which is where this fix lives, but the edit there is a single `"MIPSN32"` added to a set literal and touches no s390x path. The whole suite is 71 passed on the rebased head. The corpus, survey and cross-architecture figures were measured at `bc8b2245ff1ef9ffa5250b06972e14bd134bcf43` against `90e9094c6217e4e530bf653c06c386d8a6746f82`, and the per-test figures at `bdd5441035e02920eaa72c1c3cf9a4f0d572104d`; both are unfixed revisions and neither has been re-run at the new base, since doing so needs a native `libpyvex.so` built against this branch's `vex` pin.

zardus:

Correcting the framing in the description above, which understates this. It says
"a malformed or truncated EXRL" and gives a seven-byte truncated reproducer.
That reads as a robustness fix for bad input. It is not: **ordinary compiler
output crashes it**, and the reproducer can be a single well-formed instruction.

A backward-displacement EXRL is what compilers emit for literal-pool-relative
`MVC` and `CLC`. Six well-formed bytes are enough, and the sign of the
displacement is the whole difference:

```python
import archinfo, pyvex
pyvex.lift(bytes.fromhex("c660ffffffff"), 0x1000, archinfo.ArchS390X())   # SIGSEGV
pyvex.lift(bytes.fromhex("c66000000003"), 0x1000, archinfo.ArchS390X())   # lifts, Ijk_InvalICache
```

Both run on the released 9.3.3 wheel; the first takes the process down with no
exception to catch.

Backward EXRL is present throughout the s390x fixtures already in
`angr/binaries` — `libc.so.6`, `libstdc++.so.6`, `libm.so.6`, `ld64.so.1`,
`libgcc_s.so.1` and `test-instr_s390x` all contain them. Measured over the 337
EXRL sites in `binaries/tests/s390x`, one process per site:

| | lifted | crashed |
|---|---:|---:|
| merge base | 161 | **176** |
| this branch | **337** | **0** |

Of the 161 that already lifted, 11 produce different IR on the branch, and every
difference is a correction: the base had prefetched from between 544 bytes and
33.6 MB outside the buffer and simply had not faulted. No legitimate prefetch is
lost anywhere, and the forward-displacement control still lifts to a valid IRSB.

For whoever reviews this: **angr/vex#21**, filed in 2018, describes this failure
exactly and is referenced by neither pull request.

Two adjacent observations from the same investigation, deliberately not folded
into this change:

- The `case 0:` arm of the switch in `s390_irgen_EX` breaks without clearing
  `last_execute_target`, where every other arm clears it. The arm is selected on
  the target's top byte, so a prefetched target beginning `0x00` but not
  all-zero poisons that global for the process and silently suppresses every
  later EXRL prefetch. Not caused by this branch, but it sits next to it.
- `BRCTH` casts its 32-bit RIL-b displacement through `(Short)`, so
  `cc1600010000` lifted at `0x1000` branches to `0x1000` rather than `0x21000`.
  Any `BRCTH` beyond ±64Ki halfwords lifts to the wrong target. Separate defect,
  separate change.

zardus:

Rebased onto master at `0beff92`, because this branch had become unmergeable and was not showing it.

GitHub was answering `UNKNOWN` for this pull request's mergeability rather than `CONFLICTING`, so nothing on the page said anything was wrong. A local `git merge-tree` against master is what surfaced it.

The conflict was with #572, "Allow overriding vex_archinfo", which replaced `self.irsb.arch.vex_archinfo` with a local `vex_archinfo` throughout `LibVEXLifter.lift` — including the two lines this branch also edits, where it swaps `pvc.vex_lift` for `pvc.vex_lift_with_buffer_size`. Both changes are kept: the call now passes #572's local `vex_archinfo` *and* this branch's `buffer_size`, and the `hwcache_info` assignments on either side of the call use the local as #572 intended. `git range-diff` across the rebase shows the only alteration to the commit is that substitution; the buffer-size work is unchanged.

Not verified locally: this branch adds `vex_lift_with_buffer_size` on the C side, so exercising it needs a full pyvex build against the referenced `vex` revision, which the machine this was rebased on is not free to do right now. The rebase is a textual merge plus a build's worth of confidence from CI, not a local test run — worth saying plainly rather than implying otherwise.

zardus:

`tests/test_s390x_exrl.py` from this branch, and the three raw encodings behind it, run against the pre-fix revision this branch was cut from (pyvex `bdd5441`, vex `875f7c9`) and against this head (pyvex `ef34018`, vex `88aa12d`). Every case runs in its own child process, five times, so a fatal signal is visible and an unstable outcome cannot hide. The last row of each block is `test_s390x_exrl_real_binary`, which opens `tests/s390x/exrl/acme-client` from angr/binaries, so both sides were run with that checkout beside pyvex.

**Before** — three tests are killed by SIGSEGV, one fails its assertion every run, and one is unstable because it reads uninitialised heap past the buffer:

<details><summary>angr master</summary>

```
pyvex: .../wt/s390-b564/pyvex/__init__.py

$ pyvex.lift(bytes.fromhex("a75bffffc650ff"), 0x1000, archinfo.ArchS390X()).pp()   # aghi %r5,-1 ; truncated exrl
  [killed by SIGSEGV]
  [5 further runs: killed by SIGSEGV x5]

$ pyvex.lift(bytes.fromhex("c660ffffffff"), 0x1000, archinfo.ArchS390X()).pp()   # exrl %r6,.-2   (backward, well-formed)
  [killed by SIGSEGV]
  [5 further runs: killed by SIGSEGV x5]

$ pyvex.lift(bytes.fromhex("c66000000003"), 0x1000, archinfo.ArchS390X()).pp()   # exrl %r6,.+6   (forward, control)
  IRSB {
     t0:Ity_I64 t1:Ity_I64 t2:Ity_I64
  
     00 | ------ IMark(0x1000, 6, 0) ------
     01 | t1 = LDbe:I64(0x0000000000001006)
     02 | t4294967295 = DIRTY 1 TODO(effects) ::: s390x_dirtyhelper_EX(t1)
     03 | PUT(cmstart) = 0x0000000000001000
     04 | PUT(cmlen) = 0x0000000000000004
     NEXT: PUT(ia) = 0x0000000000001000; Ijk_InvalICache
  }
    size = 6  jumpkind = Ijk_InvalICache
  [ok]
  [5 further runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_target_before_buffer
  [5 runs: killed by SIGSEGV x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_target_past_buffer
  [5 runs: ok x4, AssertionError x1]

$ tests/test_s390x_exrl.py::test_s390x_exrl_target_past_max_bytes
  [5 runs: AssertionError x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_inside_a_larger_buffer
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_truncated
  [5 runs: killed by SIGSEGV x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_real_binary
  [5 runs: killed by SIGSEGV x5]
```

</details>

Re-measured at that same build on 2026-08-31, one fresh interpreter
per run: `test_s390x_exrl_target_past_buffer` failed its assertion in 236 runs
of 240 and passed in four, so the four passes in five recorded above are a lucky
run rather than the usual outcome. `test_s390x_exrl_real_binary` was killed in
all 40 runs of the same harness.

**After** — all seven pass, five runs out of five, and the two crashing encodings lift:

<details><summary>with this change</summary>

```
pyvex: .../wt/s390-h89/pyvex/__init__.py

$ pyvex.lift(bytes.fromhex("a75bffffc650ff"), 0x1000, archinfo.ArchS390X()).pp()   # aghi %r5,-1 ; truncated exrl
  IRSB {
     t0:Ity_I64 t1:Ity_I64 t2:Ity_I64 t3:Ity_I64 t4:Ity_I64
  
     00 | ------ IMark(0x1000, 4, 0) ------
     01 | t0 = GET:I64(r5)
     02 | t1 = Add64(t0,0xffffffffffffffff)
     03 | PUT(736) = 0x0000000000000004
     04 | PUT(744) = t0
     05 | PUT(752) = 0xffffffffffffffff
     06 | PUT(760) = 0x0000000000000000
     07 | PUT(r5) = t1
     NEXT: PUT(ia) = 0x0000000000001004; Ijk_Boring
  }
    size = 4  jumpkind = Ijk_Boring
  [ok]
  [5 further runs: ok x5]

$ pyvex.lift(bytes.fromhex("c660ffffffff"), 0x1000, archinfo.ArchS390X()).pp()   # exrl %r6,.-2   (backward, well-formed)
  IRSB {
     t0:Ity_I64 t1:Ity_I64 t2:Ity_I64
  
     00 | ------ IMark(0x1000, 6, 0) ------
     01 | t1 = LDbe:I64(0x0000000000000ffe)
     02 | t4294967295 = DIRTY 1 TODO(effects) ::: s390x_dirtyhelper_EX(t1)
     03 | PUT(cmstart) = 0x0000000000001000
     04 | PUT(cmlen) = 0x0000000000000004
     NEXT: PUT(ia) = 0x0000000000001000; Ijk_InvalICache
  }
    size = 6  jumpkind = Ijk_InvalICache
  [ok]
  [5 further runs: ok x5]

$ pyvex.lift(bytes.fromhex("c66000000003"), 0x1000, archinfo.ArchS390X()).pp()   # exrl %r6,.+6   (forward, control)
  IRSB {
     t0:Ity_I64 t1:Ity_I64 t2:Ity_I64
  
     00 | ------ IMark(0x1000, 6, 0) ------
     01 | t1 = LDbe:I64(0x0000000000001006)
     02 | t4294967295 = DIRTY 1 TODO(effects) ::: s390x_dirtyhelper_EX(t1)
     03 | PUT(cmstart) = 0x0000000000001000
     04 | PUT(cmlen) = 0x0000000000000004
     NEXT: PUT(ia) = 0x0000000000001000; Ijk_InvalICache
  }
    size = 6  jumpkind = Ijk_InvalICache
  [ok]
  [5 further runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_target_before_buffer
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_target_past_buffer
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_target_past_max_bytes
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_inside_a_larger_buffer
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_truncated
  [5 runs: ok x5]

$ tests/test_s390x_exrl.py::test_s390x_exrl_real_binary
  [5 runs: ok x5]
```

</details>

zardus:

A corpus measurement from a sweep triaging its own unowned failure classes. It
adds two things this description does not have: how often the crash fires on
real compiler output, and what happens on the displacements that do *not*
crash.

**How often.** Drawing 300 objects uniformly at random from the 7,354 s390x
objects in a private corpus of vendor- and distribution-produced binaries
(seed `segv-s390-2026-08-29`, sampled from the catalogue rather than from the
sweep's ledger, staged and sha256-verified), and probing each with `CFGFast`:
**11 die with SIGSEGV**, 267 complete, 20 fail at load, 2 were censored at a
300 s cap. That is **3.67% of s390x objects, Wilson 95% [2.06, 6.45]** —
roughly 270 of the 7,354, interval [151, 474]. The two censored objects make it
a lower bound. Scaled by the s390x share of the corpus it is 0.036% overall,
which is the honest way to read it: rare in general, common on the
architecture. None of the 11 is one of the five objects this was first found
on, so they are eleven fresh instances.

**The quiet half.** The crash is only the loud outcome. Lifting a bare six-byte
`EXRL` — so the buffer is exactly the instruction and *every* target is outside
it — on the current submodule:

```
disp     +1   target +2 bytes        -> lifts, 5 statements
disp     +4   target +8 bytes        -> lifts, 5 statements
disp    +64   target +128 bytes      -> lifts, 11 statements
disp +65536   target +131072 bytes   -> lifts, 11 statements
disp +262144  target +524288 bytes   -> lifts, 5 statements
disp     -1   target -2 bytes        -> SIGSEGV
disp   -129   target -258 bytes      -> SIGSEGV
```

Half a megabyte past a six-byte buffer, no error. And the statement count moves
with the garbage: where the out-of-buffer bytes happen to decode as an
SS-format opcode, `last_execute_target` is set and the `EXRL` is expanded
inline, so the IRSB has 11 statements instead of 5. That is silently wrong IR
built from memory the lifter was never given — which is precisely the thing
this pull request's `guest_bytes_size` exists to prevent, and it is invisible in
a crash count.

**Arms.** Two trees, each with `pyvex.__file__` asserted in-process and
`hasattr(pvc, "vex_lift_with_buffer_size")` recorded as a native-level check
that the arms really differ (False at the baseline, True here). At the
baseline, displacements -1 and -129 and a truncated encoding all take SIGSEGV
while +1 lifts; at this head with the submodule bump, all four lift cleanly. On
one 6.3 MB s390x ELF from the corpus the baseline takes SIGSEGV within seconds
under a plain `CFGFast`, and this head ran 25 minutes without crashing.

The corpus is not redistributable, so the objects are described by architecture
and format rather than named. A public reproducer needs no corpus, though: a
twelve-line Go program built `GOOS=linux GOARCH=s390x go build -buildvcs=false
-trimpath` emits 28 negative-displacement `EXRL`s against the same `MVC`
template and crashes `CFGFast` on the baseline, and it rebuilds byte-identical.
The byte-string regressions already in this branch cover the same shapes, so
that is an option rather than a gap.

session: sharpen

zardus:

A second corpus draw, seeded independently of the one above, and the half that
measurement leaves out: the same objects run against this head, not only
against the baseline.

**Arms.** Two trees materialised with `git archive` and built into separate
virtualenvs, so neither can shadow the other. Every child process records
`pyvex.__file__` and whether `pvc.vex_lift_with_buffer_size` exists, and a run
refuses to start if either disagrees with the arm it was asked for. That symbol
is this branch's own entry point, so it doubles as a native-level check that
the arms really differ. All 600 rows carry the expected arm.

- **base** — pyvex `bdd5441035e02920eaa72c1c3cf9a4f0d572104d`, this branch's
  merge base, with `vex` `875f7c9a`. Symbol absent.
- **head** — pyvex `fee0c3aa10108fad6070740705123bef370c052b`, this head, with
  `vex` `88aa12d` (angr/vex#89). Symbol present.

**Six bytes.** Each lift in its own child so a fatal signal is visible, five
runs of each case. At the merge base, `c660ffffffff` (`exrl %r6,.-2`) and
`a75bffffc650ff` (an `aghi` followed by a truncated `exrl`) are killed by
SIGSEGV in 5 of 5 runs, while the forward-displacement `c66000000003` lifts in
5 of 5. At this head all three lift in 5 of 5 — the first to `Ijk_InvalICache`,
6 bytes, 5 statements; the truncated one to `Ijk_Boring`, 4 bytes, 8 statements,
having stopped before the incomplete instruction.

**Corpus.** 300 s390x objects drawn uniformly at random from the catalogue,
seed `s390x-exrl-2026-08-29`, out of a 7,354-object s390x population inside a
757,469-object internal corpus of compiler- and vendor-produced binaries; all
300 staged and sha256-checked, none missing, all 300 probed in both arms.
`CFGFast` under a 300 s cap:

- merge base — **13 killed by SIGSEGV**, 4.33% (Wilson 95% 2.55–7.27), which
  scales to roughly 319 of the 7,354 (interval 188–535), or 0.042% corpus-wide
- this head — **0**

**What the 13 do here.** Three finish `CFGFast` outright, recovering 6,368,
8,048 and 17,282 functions. The other ten reach the 300 s cap instead — the
crash had been ending the analysis early, so removing it costs time it never
used to spend. To separate the fix from the analysis budget, the same 13 were
swept linearly through their executable sections, lifting every block, capped
at 3 MB per object: **13 of 13 SIGSEGV at the merge base; 13 of 13 complete
here, 1,639,763 blocks lifted over 32,366,812 bytes, zero lift errors.** Those
lifts all go through `vex_lift_with_buffer_size`: where angr hands over a sized
`char[]` the buffer's real extent goes with it, and where it hands over a bare
pointer the field stays zero and the bound falls back to `max_bytes`.

**Control.** 258 of the 300 completed `CFGFast` at the merge base. All 258
still complete here, 256 with identical function, block and edge counts. The
two that differ move by one function and a handful of blocks — but two runs of
the *base* arm alone on one of them give 5,506 then 5,510 functions, so
`CFGFast` is not reproducible at that size and neither difference is
attributable to this change. The 22 objects that never reach the CFG at all — 18
to an `AssertionError` in pyelftools' `dynamic.py:183:_get_stringtable`,
reached through `cle`'s ELF backend, and 4 to a `NotImplementedError` at angr's
`sim_type.py:4455`, both inside `Project.__init__` — fail identically on both
arms.

**Four of the 300 are public.** They are byte-identical to files tracked in
`angr/binaries` at `9d9d9e0` — `tests/s390x/allcmps`, `tests/s390x/ctype_b_loc`,
`tests/s390x/scanf_test`, `tests/s390x/vfg_0` — and give the same CFG on both
arms, so they are control rather than reproducers.

The rest of the corpus is not redistributable, so those objects are described
by architecture, format and sha256 rather than named.

session: sharpen

ltfish:
Quick check: Does this bug exist in the upstream libvex / valgrind code?

I strongly prefer that we do not pass a second size parameter to the `lift()` method unless necessary.

zardus:

No, it is ours. Upstream reads the target from the guest address, not from the code buffer handed to the translator, so the fork's index into that buffer has no counterpart there. `exrl_bytes` does not appear anywhere in the upstream file.

```c
/* valgrind c7c8188, VEX/priv/guest_s390_toIR.c:11470 */
bytes_addr = addr_rel_long(offset);
bytes = (UChar *)(HWord)bytes_addr;
/* angr/vex 875f7c9, priv/guest_s390_toIR.c:13064 */
UChar *bytes = exrl_bytes + offset * 2UL;
```

On the second size, I think it is necessary: the case below faults without it and does not with it. A 1000-byte buffer ending flush at a `PROT_NONE` page, `bytes_offset=900`, `max_bytes=1000`, and an EXRL whose target is 500 bytes on, so 100 bytes are readable and `guest_max_bytes` is 1000. With the extent stated the branch refuses the target and returns a 6-byte `Ijk_InvalICache` block; with it unstated the read takes SIGSEGV. That is a plain `pyvex.lift()` call, but I could not find a caller who makes it -- angr's `_load_bytes` clamps the size to `len(backer) - offset`, and cle lifts Python bytes at offset 0 or 1 -- so it hardens a hole that is reachable through the public API and that nothing in angr, cle or pyvex reaches today. A bare pointer leaves the extent unstated after this merges too, so the same overstated bound still faults there.

What it changes today is IR. At angr's 400-byte cap, over `binaries/tests/s390x` at master `981989f` with the whole file as the lift buffer, 101 of 1310 EXRL sites stop resolving their target without the extent, and the block ends at the EXRL instead of carrying on.

I kept the argument off `vex_lift` by putting `vex_lift_with_buffer_size` beside it, so no signature changed -- `pyvex.lift()` and `Lifter.lift()` included.

session: sharpen

zardus:

Closing: this is the pyvex half of angr/vex#89, which is closed as superseded.

vex master's Valgrind 3.27.1 import carries an `#ifdef PYVEX` clamp in `s390_irgen_EXRL` that keeps the EXRL prefetch inside the lift buffer, and pyvex master already sets `guest_max_bytes` to the real buffer size on every lift (`pyvex_c/pyvex.c:369`, `LibVEX_set_VexControl` at :390) -- which is the job this branch's plumbing was doing. The `guest_bytes_size` field is only needed for the wider reach vex#89 wanted, so it goes with it.

Not lost: the three extra EXRL tests this branch adds to `tests/test_s390x_exrl.py` are regression cover for master's own clamp, and the acme-client fixture they read is already on angr/binaries master under `tests/s390x/exrl/`. They will be re-proposed as a small pyvex-only pull request, with the assertions updated for the IR master's `s390_irgen_EX` now emits.

Paired 1:1 with angr/vex#89.

PR #568, opened 2026-08-20
Lifting: Handle mixed-width Thumb IT lookback

zardus:

### Problem

Lifting `mixed_width_after` at `0x110c6` in the compiler-built [Thumb fixture](https://github.com/angr/binaries/blob/97ce07863c6d80a4145b73201dcb80d2da583332/tests/armel/thumb_it_lookback_widths.bin) retains ITSTATE after its mixed-width IT group has ended. At optimization level 1, the baseline emits a false conditional exit to Thumb address `0x110c9`.

### Root cause

VEX's lookback allowed two halfwords per guarded instruction instead of counting actual Thumb widths. It could therefore treat an expired IT as active, and a false `0xbf08` candidate inside a 32-bit `LDREX` could stop the search before an older real IT.

### Fix

Adopt the VEX decoder change that walks actual Thumb widths and continues past expired candidates. Replace synthetic regression bytes with the synchronized compiler-built Cortex-M3 fixture and make the direct and Pyodide test jobs resolve it.

### Testing

The compiler-built Thumb fixture covers mixed-width expiry at optimization levels 0–2, conservative insufficient or disabled lookback, an expired false candidate inside `LDREX`, and the nine-halfword window edge. Validation: https://github.com/angr/pyvex/pull/568#issuecomment-5361007401

The VEX dependency must land before this submodule update; the fixture dependency supplies the regression input.

sync: angr/binaries#212
sync: angr/vex#92

session: sharpen


zardus:

Validation record for head `d4abbd0385f59198f4b6d2225bd03aa15c0b4dc9` against baseline `24e2bd862aaedc14887bd247f1f1130445f94db6`.

Coordinated candidate: PyVEX `d4abbd0385f59198f4b6d2225bd03aa15c0b4dc9` with VEX `866bb8c991ed59fe37352443fbadbf9bef160d5a`; baseline: PyVEX `24e2bd862aaedc14887bd247f1f1130445f94db6` with VEX `e3062871112afded52ae155757113a5873e9819c`. Tests used CPython 3.12.13 and the built packages.

Rebase: the ARM source blob and the PyVEX regression/workflow blobs are unchanged from the preceding published heads. The PyVEX gitlink now names the VEX rebased commit, preserving the AVX512 baseline.

- Native build: the combined VEX/PyVEX store build passed, including the rebased ARM decoder and the AVX512 changes in the VEX baseline.
- Focused regression: `pytest --import-mode=append -q tests/test_arm_postprocess.py -k thumb_it_lookback` against the baseline produced 1 failed, 3 passed, 2 deselected; the candidate produced 4 passed, 2 deselected. The failing baseline test is `test_thumb_it_lookback_mixed_width_group_ends_before_block` (ITSTATE remains live). The same compiler-built fixture was used in both runs.
- Full PyVEX suite: `pytest --import-mode=append -q tests` passed 91 tests on the candidate.
- Efficacy at optimization level 1: `mixed_width_after` changes from one ITSTATE read and one false exit to `0x110c9`, to no ITSTATE reads or exits. The older-candidate control retains its exit to `0x110dd`; the nine-halfword window-edge control remains unconditional.
- Fixture: binaries commit `97ce07863c6d80a4145b73201dcb80d2da583332`, `tests/armel/thumb_it_lookback_widths.bin`, 64 bytes, SHA-256 `a49c1a151765f87446da09b98d6b11991f054a0a4ccb1d4ed2ea92694ad1f628`.
- Lint/type comparison for the changed Python file: Pylint 10.00 to 10.00; Pyright errors 1 to 1. Configured PyVEX pre-commit hooks, test-input policy, and test-package checks passed.

### Workspace gate and limits

`./feature.sh test pr-pyvex-568` was interrupted during feature-build. Before interruption, workspace tests reported a manager test fixture error (`NOT NULL constraint failed: pr.repo`); that test passes on workspace commit `f0587f442a330e471c7c3a0b58f27ba418f54cb2`, whose fixture supplies the missing PR row. Test-inputs, test-packages, and pre-commit completed successfully.

The remaining suites ran with `./feature.sh test pr-pyvex-568 feature-build mono pyvex pysoot`: mono, PyVEX, and PySoot passed. Feature-build failed while preparing sweep tools because Nix reported an invalid Python build-tool store reference. The gate exited 1; this is not a full green workspace gate.

Suites skipped because their repositories were not adopted: archinfo, pypcode, cle, angr, angr-rust, and angr-management. Workspace, test-inputs, test-packages, and pre-commit were not repeated in the resumed command. Decompiler snapshots were not run locally.

This replaces the earlier-head validation. Its corpus, fixture-rebuild, mutant-control, and hosted-CI measurements were not rerun for these heads and are not claimed as current validation.

### Fixture dependency refresh

Run [35043552714, attempt 1](https://github.com/angr/pyvex/actions/runs/35043552714/attempts/1) built binaries `97ce07863c6d80a4145b73201dcb80d2da583332`. Failing shards 0, 1, 2, 3, 4, 6, and 9 report missing fixtures from subsequent binaries commits. The synchronized dependency has been rebased onto `496378f5cb30bb057e0d5bf76637b4e2c3c224d5`, producing `7028ab7a8035926d0fe92fd41ceb125dd5fa19ea`. Its Thumb fixture, source, and recipe are byte-identical; the 16 identified missing fixture paths are present with blobs identical to that baseline.

All 91 PyVEX tests pass with the refreshed fixture checkout. PyVEX and VEX source heads are unchanged. The whole hosted workflow needs a fresh Build archive; no hosted success after the dependency refresh is claimed here.

The completed gate `./feature.sh test pr-pyvex-568` exited 0: PARTIAL PASS (8 of 14 suites). Workspace, test-inputs, test-packages, pre-commit, feature-build, mono, PyVEX, and PySoot passed. Suites skipped because their repositories were not adopted: archinfo, pypcode, cle, angr, angr-rust, and angr-management. The gate does not run hosted decompiler snapshots or the full unadopted ecosystem suites.

zardus:

Correcting my earlier comment on this PR. It was wrong in several ways: it said the guards "never exist" under this change, it mispaired one of the two functions it cited with a level and binary where that function does not move, it gave a count of GED regressions its own evidence does not support, and it claimed the effect had been isolated to this PR when it had not. The measurement below replaces the argument rather than patching it.

**What prompted the look, stated accurately this time.** Some libopencm3 functions scored worse on structural GED against source when an [angr/vibr](https://github.com/angr/vibr) preview build was compared against master. That build carried several changes, so nothing in those scores is attributable to this PR alone — my earlier "confirmed by swapping only pyvex" claimed an isolation the experiment did not have. I can no longer re-run that harness, so I am not restating its tally either. Two of the functions it scored worse, `gpio_set_af` (45 of the measured entries below) and `gfx_drawChar` (3), turn out to lift **identically on both arms** of the clean A/B below, so at least some of those regressions cannot be this change at all.

**A clean A/B, differing only in pyvex and its VEX submodule.**

- A: pyvex `dbc5b170ca6ccc7fa688da25a9d8d09b3d73d8e3`, VEX `875f7c9a5f6be621b4f000c29c016e15ddf32207`
- B: pyvex `86716e788dc944bda3c0f7000096a15c1d7783ef` (this PR), VEX `881703eda064a67a803cf51974c992286a8bdf4f` (angr/vex#92)

`dbc5b170` is an ancestor of this PR's base `ab18a834ba4a5320c3415eff971092ef2b940779`, three commits behind it. The two are interchangeable for this measurement: `git ls-tree <rev> vex` is `875f7c9a` at both, and the only Python change between them is an optional `vex_archinfo=None` parameter whose default resolves to the pre-existing `arch.vex_archinfo`.

Corpus: the 69 libopencm3 ELFs in the public [noelo-lab/decbench-dataset](https://huggingface.co/datasets/noelo-lab/decbench-dataset) HuggingFace dataset at revision `e5eb576d66ee36793b800a4dd45e291e0add4472` (BSD-2-Clause) — 23 example programs at O0, O2 and O2-noinline. All 69 are byte-identical to the copies used below.

Lifting the body of every `STT_FUNC` symbol of nonzero size, block by block, on both arms. The corpus holds 11,506 `STT_FUNC` symbols, 51 of them zero-sized and therefore skipped, leaving 11,455; those cover only 4,399 distinct `(level, binary, address)` code entries, because the Cortex-M weak vector table puts up to 96 handler names on one address. Aliases at an address carry identical counts, so the table counts each distinct entry once:

| counter | A | B |
| --- | --- | --- |
| `armg_calculate_condition` CCalls | 2,325 | 1,305 |
| `ITSTATE` reads | 816 | 222 |
| `Ist_Exit` statements | 18,259 | 17,898 |
| `Iex_ITE` | 8,685 | 6,387 |

255 entries differ between the arms — 64 distinct symbol names, 141 distinct `(name, level)` shapes, across 12 of the 23 programs. **Every one moves down on at least one counter and none moves up on any counter; zero increases anywhere in the corpus.**

Where guards are removed outright rather than reduced: of the 498 entries carrying `armg_calculate_condition` CCalls on A, **89 have none on B**; of the 271 carrying `ITSTATE` reads on A, **176 have none on B**.

B's reachable block set is a subset of A's in every entry: 38,240 `(entry, block)` pairs are reached by both, 141 by A only, none by B only. Each of the 134 entries holding those 141 blocks loses at least one `Ist_Exit` on the blocks both arms reach, which is consistent with them being successors of exits this change proves dead — the probe records counts, not successor addresses, so that last step is inference rather than a per-block proof. No IRSB changes size on the shared blocks.

Counted per symbol instead, with every alias separately, the same corpus gives 3,441 → 1,761 CCalls, 1,248 → 222 `ITSTATE` reads, 20,131 → 19,626 `Ist_Exit` and 14,205 → 10,251 ITEs over 11,455 symbols, 411 of which change. Those totals are over all blocks each arm reached, not over the shared set. Counted per symbol the block figures become 225 A-only and 49,775 shared. No increases under either counting.

**The two functions the earlier comment named**, at the level it quoted for each. Neither is emptied of guards, and one of them does not move at all in the binary that was cited:

- `usb_control_recv_chunk` at O2-noinline, the level quoted for it: 11 → 5 CCalls and 15 → 7 ITEs, with `ITSTATE` reads and `Ist_Exit` both unchanged at 2 — identical in all three binaries carrying it. At O2 the same function is 15 → 5 CCalls and 33 → 7 ITEs.
- `usb_control_request_dispatch` at O2-noinline in `msc.elf`: 0 CCalls, 0 `ITSTATE` reads and 0 ITEs on **both** arms, so there was nothing there to change — this is the pairing my earlier comment got wrong. It does move at that level in `cdcacm.elf` and `usbmidi.elf` (3 → 2 CCalls, 5 → 3 ITEs), and its largest move is at O2 in `msc.elf`: 9 → 4 CCalls, 3 → 1 `ITSTATE` reads, 9 → 5 `Ist_Exit`.

So for these two the guards shrink rather than disappear, which is what my "the guards never exist" sentence got wrong.

**Reproducing it.** Build both pyvex/VEX pairs above and fetch the dataset at the pinned revision. For each ELF and each `STT_FUNC` symbol of nonzero size, walk the symbol's own body as a worklist of blocks: start at `st_value & ~1`, stay inside `[st_value & ~1, +st_size)`, and follow each `Ist_Exit`'s constant target, the constant `irsb.next`, and the fall-through after a call. Per block, count `Iex_CCall`s to `armg_calculate_condition`, `Iex_Get`/`Iex_GetI` at the ARM guest `ITSTATE` offset (392), `Iex_ITE`s and `Ist_Exit`s, and record `irsb.size`.

One detail decides whether the numbers reproduce: pass the **whole section's bytes** to `pyvex.lift` with `bytes_offset = (addr | 1) - section_vaddr` for Thumb, arch `ARCH_ARM_LE`, `opt_level=1`. A pre-sliced buffer, or an offset computed without the Thumb bit, makes libVEX decode one byte early and silently return a smaller IRSB — no error, just wrong figures. I can attach the script.

Two caveats on the counters: `Ist_Exit` is counted as a statement, so this does not prove each remaining one has a non-constant guard; and the per-symbol totals count a block once per symbol that reaches it.

Nothing here asks for a change to the PR; it corrects what this comment claimed about it.

zardus:

Lift result for `mixed_width_after` at `0x110c6` in the [compiler-built Thumb fixture](https://github.com/angr/binaries/blob/97ce07863c6d80a4145b73201dcb80d2da583332/tests/armel/thumb_it_lookback_widths.bin), at optimization level 1.

**Before**

<details><summary>pyvex baseline 24e2bd862aaedc14887bd247f1f1130445f94db6</summary>

```json
{"case": "mixed_width_after", "itstate_reads": 1, "exits": ["0x110c9"], "size": 2, "jumpkind": "Ijk_Ret"}
```

</details>

**After**

<details><summary>pyvex head d4abbd0385f59198f4b6d2225bd03aa15c0b4dc9</summary>

```json
{"case": "mixed_width_after", "itstate_reads": 0, "exits": [], "size": 2, "jumpkind": "Ijk_Ret"}
```

</details>

To reproduce, read fixture bytes `[0:0x14]` and construct `pyvex.IRSB(data, 0x110c7, pyvex.ARCH_ARM_LE, max_bytes=2, num_inst=1, bytes_offset=0x13, opt_level=1)`. Count reads at the architecture's `itstate` offset and list `irsb.exit_statements` targets. Both builds use the same fixture bytes.

PR #570, opened 2026-08-23
Lifting: Resolve s390x BRCTH branch targets past a 16-bit displacement

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

Lifting the first instruction in `tests/s390x/brcth-displacements.bin` at `0x1000` gives a taken target of `0x1000` instead of `0x21000`. The BRCTH displacement is +65536 halfwords, so truncating it produces a spurious self-loop. The large negative case is truncated too.

### Root cause

`s390_irgen_BRCTH` passes its 32-bit displacement to `addr_relative(UShort)`, which narrows it to 16 bits.

### Fix

Update the VEX submodule to the correction in angr/vex#93, which uses `addr_rel_long(UInt)` to sign-extend the complete field and calculate the relative address.

### Testing

`tests/test_s390x_brcth.py` loads assembled instructions and checks the taken target, fallthrough, size and jumpkind for large positive and negative displacements and small-offset controls. The large cases fail against the baseline. The platform and Pyodide jobs now fetch the fixture through the binaries reference resolver.

The fixture is supplied by (angr/binaries#234). Merge that fixture and angr/vex#93 before this PR.

Validation: https://github.com/angr/pyvex/pull/570#issuecomment-5384413863

sync: angr/binaries#234

session: sharpen

PR #571, opened 2026-08-23
Lifting: Keep a stale s390x execute target from suppressing later EXRL expansion

zardus:

### Problem

Lifting an s390x `EXRL` whose six execute-target bytes begin `0x00` silently degrades every later `EXRL` in the same process. With the vectors in `tests/test_s390x_ex_stale_target.py`, an expandable `EXRL` expands its execute target on a clean process, stops expanding after one such block is lifted, and lifts to an opaque invalidate-and-restart block from then on.

One block of misidentified code therefore corrupts the lift of every s390x binary analysed afterwards, and nothing reports it.

### Root cause

libVEX caches the execute target in a global. One arm of `s390_irgen_EX` — the arm selected when the target's first byte carries no code information yet — returns without clearing it, so the stale value outlives the lift.

### Fix

Clear the global on that arm too. The fix does not depend on the state the process starts in, since the global is cleared whichever way the arm is entered.

### Testing

`tests/test_s390x_ex_stale_target.py` asserts across two lifts, because the state being tested is a libVEX global rather than anything reachable from one IRSB. It fails against the baseline submodule. With the file present, `tests/test_s390x_exrl.py::test_s390x_exrl` also fails on the baseline, because the stale global survives into the following test file. Needs angr/vex#94, which the submodule bump here already points at. The sequence is in the comment below.

The submodule bump conflicts textually with the ones in #564 and #570, so whichever lands first, the others rebase.

Validation: https://github.com/angr/pyvex/pull/571#issuecomment-5384416632

Merge order: angr/vex#94 first. This head's `vex` submodule is `f7b8fcc`, which is that pull request's head and one commit ahead of `vex` master, so merging this half on its own leaves pyvex master's submodule pointing at a commit that is on no vex branch. `angr/vex` is not in the CI repository list, so the `sync:` line below is a note to the reader rather than an instruction to CI.

sync: angr/vex#94

session: sharpen

### DIFF
--- a/tests/test_s390x_ex_stale_target.py
+++ b/tests/test_s390x_ex_stale_target.py
@@ -1,0 +1,22 @@
+import pyvex
+
+# exrl %r1,0x400408; br %r14; xc 0(0,%r2),0(%r3); padding
+EXPANDABLE = b"\xc6\x10\x00\x00\x00\x04\x07\xfe\xd7\x00\x20\x00\x30\x00\x7d\xa7"
+
+# The same block with an execute target whose first byte is 0x00. libVEX
+# selects the arm of s390_irgen_EX that has no code information yet on that
+# byte, and caches the target in a global that outlives the lift.
+LEADING_ZERO_TARGET = b"\xc6\x10\x00\x00\x00\x04\x07\xfe\x00\x07\x20\x00\x30\x00\x7d\xa7"
+
+
+def expands_execute_target(data):
+    return "0xd700200030000000" in str(pyvex.lift(data, 0x400400, pyvex.ARCH_S390X))
+
+
+def test_s390x_ex_leading_zero_target_does_not_persist():
+    pyvex.lift(LEADING_ZERO_TARGET, 0x400400, pyvex.ARCH_S390X)
+    assert expands_execute_target(EXPANDABLE)
+
+
+if __name__ == "__main__":
+    test_s390x_ex_leading_zero_target_does_not_persist()


zardus:

Validation record for head `aff4c7465de1e485a0c26ee83f20341e6a9d8aff` against baseline `ab18a834ba4a5320c3415eff971092ef2b940779`.

- Regression: `pytest tests/test_s390x_ex_stale_target.py` — fails against the baseline submodule (after lifting an `EXRL` whose target begins 0x00, the next lift of an expandable `EXRL` no longer expands), passes against the bumped one
- Blast radius on the baseline: with this file present, `tests/test_s390x_exrl.py::test_s390x_exrl` also fails, because the stale global survives into the following test file
- Full suite: `pytest tests` — 65 passed, 0 failed, 0 skipped, on Python 3.12.13 / x86-64 Linux
- Lint: `pre-commit run --files tests/test_s390x_ex_stale_target.py` — all hooks pass, tree unchanged
- vex build: `make -f Makefile-gcc -j8 all` at `f7b8fccd938be070e4a1d53692f9f16669d66627` — clean
- Workspace gate: not run — the shared environment is pinned to a live corpus sweep, so validation used a separate checkout and virtualenv. Only this repository's suite ran; angr, cle, claripy and the GUI suites did not.

The defect survives #564: its bounds check still admits an in-buffer target, and the sixteen bytes `c6100000000407fe000720003000` plus padding reproduce it there too.

Caveats: the regression asserts across two lifts because the state is a libVEX global; it does not depend on the state the process starts in, since the fix clears the global whichever way that arm is entered.

Re-validated on this head after rebasing onto the baseline above: the focused suites above re-run green (1 passed).

Re-validated on this head after rebasing onto the baseline above: the focused suites above re-run green (1 passed).

zardus:

The defect is stateful, so the evidence is a sequence rather than one lift: the same expandable `EXRL` is lifted, then an `EXRL` whose execute target begins `0x00` is lifted, then the first one again. Vectors are the ones in `tests/test_s390x_ex_stale_target.py`.

**Before** — the second lift of the identical block no longer expands its execute target. One block of misidentified code has silently degraded every later `EXRL` in the process:

<details><summary>pyvex master, vex 875f7c9</summary>

```
--- an expandable EXRL, lifted first in a clean process
    execute target expanded: True
--- now lift an EXRL whose execute target begins 0x00
    lifted
--- the same expandable EXRL, lifted again afterwards
    execute target expanded: False
```

</details>

**After** — the global is cleared on that arm too, so the third lift behaves like the first:

<details><summary>with this change, vex f7b8fcc</summary>

```
--- an expandable EXRL, lifted first in a clean process
    execute target expanded: True
--- now lift an EXRL whose execute target begins 0x00
    lifted
--- the same expandable EXRL, lifted again afterwards
    execute target expanded: True
```

</details>

Reproduce with `nix develop --command python scratch/outputs/repro/pyvex_571.py`.

zardus:

Closing: this is the pyvex half of angr/vex#94, which is closed as superseded.

vex master's Valgrind 3.27.1 import replaced the `case 0` arm of `s390_irgen_EX` with an explicit `Invalid_execute_target` sentinel and resets to it on both paths, so the stale execute target this branch tests for cannot persist. Apart from the `vex` submodule bump the branch is a 22-line test built from hand-assembled bytes, and it is not being carried forward.

Paired 1:1 with angr/vex#94.

